Issue #56, opened 2026-06-24
Reidentification_net training issue RuntimeError: mat1 and mat2 must have the same dtype, but got Half and Float


jugal-sheth:
**RuntimeError: mat1 and mat2 must have the same dtype, but got Half and Float**

Issue
Training with with_center_loss: True and AMP/mixed-precision enabled fails with:
RuntimeError: mat1 and mat2 must have the same dtype, but got Half and Float

**Steps to reproduce bug**
while training reid under specs change following:

`1. metric_loss_type: triplet_center`
`2. with_center_loss: True`

**Expected behavior**

Model Trains as normal

**Environment overview**

 - Environment location: docker
 - Method of TAO Toolkit Installation install: Getting started Notebook
 - TAO command
`!tao model re_identification train \
                  -e $SPECS_DIR/experiment_VesselReID_resnet.yaml \
                  results_dir=$RESULTS_DIR/VesselReID \
                  encryption_key=$KEY` 

**Environment details**
docker image ` nvcr.io/nvidia/tao/tao-toolkit:6.26.3-pyt`

jugal-sheth:
https://github.com/NVIDIA-TAO/tao-pytorch/pull/57

PR #57, opened 2026-06-25
Fix AMP dtype mismatch in CenterLoss.forward (Half vs Float)

jugal-sheth:
# What does this PR do ?

1. Fix for [issue ](https://github.com/NVIDIA-TAO/tao-pytorch/issues/56) #56 
2. Fix minor spelling mistake.

# Issue #
Training with with_center_loss: True and AMP/mixed-precision enabled fails with:
`RuntimeError: mat1 and mat2 must have the same dtype, but got Half and Float`
`Traceback points to center_loss.py, line 80:`

`distmat.addmm_(x, self.centers.t(), beta=1, alpha=-2)`

Under torch.cuda.amp.autocast, the input feature tensor x is cast to float16 (Half) by autocast's standard ops, but self.centers is an nn.Parameter that remains float32. The elementwise + used to build distmat implicitly upcasts to float32, so distmat ends up float32 — but addmm_ is an in-place op, which autocast does not wrap/cast automatically. The result is distmat (float32) receiving x (float16) and self.centers.t() (float32) as operands to the same addmm_ call, causing the dtype mismatch.
This breaks any AMP training run that uses center loss (metric_loss_type: triplet_center with with_center_loss: True), forcing users to either disable AMP or disable center loss to avoid the crash — both of which reduce training quality/speed unnecessarily.

# Fix #
Cast x to self.centers.dtype at the start of forward, before any computation. This keeps centers (the accumulated learnable parameter) in float32 for precision, while ensuring all tensors entering addmm_ share a consistent dtype.

**Collection**: 
ReIdentificationNet

# Changelog 
Modified `"tao-pytorch\nvidia_tao_pytorch\cv\re_identification\model\losses\center_loss.py"`
- `# Cast to Centers' to avoid AMP autocast Half/Float mismatch`
       ` x = x.to(self.centers.dtype)`

# Before your PR is "Ready for review"
**Pre checks**:
- [X] Make sure you read and followed [Contributor guidelines]
- [X] Did you write any new necessary tests?
- [X] Did you add or update any necessary documentation?
- [X] Does the PR affect components that are optional to install? (Ex: Numba, Pynini etc)
- [x] Reviewer: Does the PR have correct import guards for all optional libraries?
  
**PR Type**:
- [ ] New Feature
- [X] Bugfix
- [ ] Documentation

If you haven't finished some of the above items you can still open "Draft" PR.

## Who can review?

Anyone in the community is free to review the PR once the checks have passed.


hqtran-nv:
Thanks for this @jugal-sheth , and thanks for the clear write-up in the description.

The analysis is correct. `precision='16-mixed'` is set unconditionally for re-identification training, so any run with `with_center_loss: True` hits this — the only reason it went unnoticed is that the option defaults to `False`. Casting `x` up to the centers' dtype is the right direction, since it keeps the accumulated centers in fp32 rather than degrading them to half, and the non-AMP path is unaffected. The change also covers both `center` and `triplet_center` loss types, since it sits inside `forward`.

Two things before this can be merged:

1. Please merge the latest `main` into your branch. It has moved on since you opened this, and a refresh will let the checks run cleanly.
2. Please add yourself to `CONTRIBUTORS.md` in the same PR. The contributor table is empty, so you would be the first name on it.

Your existing commit is signed off correctly; please keep `-s` on the new ones so the DCO check stays green.

Once those are in I will run CI and get this merged. Thanks for the fix and for the patience on the review.

jugal-sheth:
Thanks @hqtran-nv  Glad to be a contributor, and looking forward to getting this merged soon!

hqtran-nv (approved):

hqtran-nv:
Hi @jugal-sheth ,

Your PR has been merged.

Thank you for your contribution.

PR #67, opened 2026-08-05
[TAO-0000][Bugfix] Fix AutoML checkpoint resume behavior

ramanathan831:
## Summary

Migrates the PyTorch framework changes and regression coverage from the discontinued GitLab AutoML MR.

## Changes

- makes resumed Sparse4D training consume the promoted step budget as the sole stopping condition
- adds regression coverage for the Sparse4D resume budget
- adds BEVFusion CPU rotated-IoU fallback coverage

## Root cause

Sparse4D resumed runs could stop on the inherited epoch limit before consuming the promoted step budget. BEVFusion also needed explicit coverage that its default rotated-IoU path remains CPU-safe and does not initialize CUDA.

## Validation

- patches replayed cleanly against GitHub `main`
- local collection reached TAO-only dependencies; authoritative execution is delegated to repository CI because the host lacks `mmdet3d` and the full TAO test environment


nvidia_tao_pytorch/cv/sparse4d/scripts/train.py
```diff
@@ -72,6 +72,13 @@ def run_experiment(experiment_config, key):
         callbacks.append(PklResampleCallback(num_iters_per_epoch=num_iters_per_epoch))
         logging.info(f"PklResampleCallback enabled (interval={num_iters_per_epoch} steps)")
 
+    # Sparse4D defines its training budget in optimizer steps.  Keeping the
+    # common max_epochs limit as well can terminate a resumed run at the epoch
+    # boundary encoded in the checkpoint before the increased max_steps budget
+    # is consumed.  Let max_steps be the single authoritative stop condition so
+    # multi-fidelity AutoML promotions actually execute their additional steps.
+    trainer_kwargs["max_epochs"] = -1
```
vpraveen-nv, line 85: This applies to *all* sparse4d runs, not just AutoML resumes — which is fine today because `max_steps` is always passed to `Trainer(...)` a few lines down. But it makes `max_steps` load-bearing in a non-obvious way; if someone later makes `max_steps` conditional, training becomes unbounded. A one-line guard or comment tying the two together (e.g. "safe only because max_steps is always set below") would prevent that footgun.

tests/cv_unit_test/sparse4d/test_trainer.py
```diff
@@ -101,6 +102,46 @@ def test_trainer_fit(_train_spec):
     trainer.fit(pt_model, dm)
 
 
+def test_run_experiment_uses_step_budget_as_only_stop_condition(monkeypatch):
```
vpraveen-nv, line 107: Nit: the neighboring tests in this file carry `@pytest.mark.cv_unit` / `@pytest.mark.sparse4d`; this one has neither, so marker-based selection (`pytest -m "cv_unit and not tensorrt"`) will skip it. Add the same markers.

tests/pointcloud_unit_tests/bevfusion/test_rotate_iou.py
```diff
@@ -0,0 +1,45 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression tests for the BEVFusion CPU rotated-IoU fallback."""
```
vpraveen-nv: Scope question: this file tests the BEVFusion rotated-IoU CPU fallback, which is unrelated to the PR title (AutoML checkpoint resume). The tests themselves look correct against the module on `main`, but either mention them in the PR description or split them out so the change history stays greppable. Also consider a `@pytest.mark.bevfusion`-style marker consistent with the other pointcloud tests.

vpraveen-nv (commented):
minor nits.

nvidia_tao_pytorch/cv/sparse4d/scripts/train.py
ramanathan831: Fixed in c1246636; max_steps remains the sole stop condition and this is documented next to max_epochs=-1.

tests/cv_unit_test/sparse4d/test_trainer.py
ramanathan831: Fixed in 8a026fb8: added cv_unit and sparse4d markers.

tests/pointcloud_unit_tests/bevfusion/test_rotate_iou.py
ramanathan831: Fixed in 8a026fb8: added cv_unit and bevfusion markers; the focused CPU fallback regression remains in this test PR.

vpraveen-nv (approved):
All comments addressed — the step-budget guard + comment, the test markers, and the bevfusion markers are all in. Verified `max_steps` remains always-set. LGTM.

PR #68, opened 2026-08-05
[TAO-0000][Bugfix] Correct Mask Grounding DINO COCO evaluation

ramanathan831:
## Summary
- port the Mask Grounding DINO COCO evaluator fix
- include the sealed validation overlay used by the original GitLab MR

## Root cause
Evaluator behavior and its validation overlay existed only in the discontinued GitLab branch.

## Validation
- branch rebased onto current GitHub `main`
- `git diff --check`


release/mask_grounding_dino_coco_evaluator_overlay/build_overlay.py
```diff
@@ -0,0 +1,145 @@
...
+import hashlib
+import io
+import json
+from pathlib import Path
+import subprocess
+import tarfile
+
+
+BASE_REF = "origin/release/7.1.0"
+BASE_COMMIT = "99741bc8229617d0d3dd52e30540111d55efd1af"
```
vpraveen-nv: This overlay tooling doesn't belong on `main`: (1) `_base_blob` does `git show 99741bc8…:path`, but that commit comes from GitLab-era `release/7.1.0` — GitHub `main` is a squashed migration root with no shared history, so the script fails in any GitHub clone that doesn't also fetch the legacy ref; (2) `CONTAINER["sqsh_path"]` hardcodes a personal Lustre path (`/lustre/fsw/portfolios/edgeai/users/rarunachalam/...`), violating the no-hardcoded-paths rule; (3) the manifest pins the current `pl_gdino_model.py` bytes against a 7.1.0 base, so it goes stale the moment anyone touches that file, and `release/` is otherwise reserved for packaging. Link the validation evidence from the PR description instead, and drop the directory from the merge.

tests/cv_unit_test/mask_grounding_dino/test_coco_evaluator_routing.py
```diff
@@ -0,0 +1,80 @@
...
+        if not isinstance(node, ast.Call):
+            continue
+        if isinstance(node.func, ast.Name):
+            names.append(node.func.id)
+        elif isinstance(node.func, ast.Attribute):
+            names.append(node.func.attr)
+    return names
+
+
+def test_object_detection_uses_distributed_coco_evaluator():
```
vpraveen-nv, line 53: This file has no pytest markers. The unit suites are selected by marker (e.g. `pytest -m "cv_unit and not tensorrt"`), so without `@pytest.mark.cv_unit` these tests are silently deselected — defeating the point of adding regression coverage. Add the marker, matching the other tests in `tests/cv_unit_test/mask_grounding_dino/`.

tests/cv_unit_test/mask_grounding_dino/test_coco_evaluator_routing.py
```diff
@@ -0,0 +1,80 @@
...
+        assert "synchronize_between_processes" in calls
+        assert "overall_accumulate" in calls
+        assert "overall_summarize" in calls
+
+
+def test_object_detection_emits_task_correct_bbox_and_mask_metrics():
+    source = MODEL_PATH.read_text(encoding="utf-8")
+
+    for stage in ("val", "test"):
+        assert f'"[{chr(123)}iou_type{chr(125)}] {stage}_{chr(123)}key{chr(125)}"' in source
```
vpraveen-nv: This assertion greps the raw source for an exact f-string literal (using `chr(123)`/`chr(125)` to spell braces), and the AST test above pins call names. Both couple to formatting rather than behavior: a harmless refactor breaks them while a functionally broken evaluator (wrong `stats` index, wrong constructor args) still passes. Prefer a real unit test: drive `on_validation_epoch_end` with a stubbed `CocoEvaluator` and assert the logged keys and that sync/accumulate/summarize ran on all ranks. At minimum drop the `chr()` obfuscation — `'"[{iou_type}] ' + stage + '_{key}"'` reads the same without the puzzle.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
```diff
@@ -369,16 +368,25 @@ def on_validation_epoch_end(self):
...
+            self.val_evaluator.overall_summarize(is_print=False)
+            if self.trainer.is_global_zero:
                 for iou_type in self.iou_types:
-                    # _summarize_task omits the "all" aggregate when no class produced
-                    # predictions (e.g. fast_dev_run with an undertrained model).
-                    for key, value in eval_results[iou_type].get('all', {}).items():
-                        if isinstance(value, (int, float)):
-                            self.log(f"[{iou_type}] val_{key}", value, rank_zero_only=True)
-                        self.status_logging_dict[f"[{iou_type}] val_{key}"] = str(value)
+                    stats = self.val_evaluator.coco_eval[iou_type].stats
```
vpraveen-nv, line 401: The status.json KPI / logged-metric names change here: old code emitted the `OD_Evaluator` per-key names, the new code emits `"[bbox] val_mAP@50-95"` / `"[bbox] val_mAP@50"` (and `[segm]` variants). Two checks: (1) do any downstream consumers — AutoML monitor / `train.optim.monitor_name`, docs, notebooks — reference the old names? (2) sibling grounding_dino uses plain `val_mAP` / `val_mAP50` for the same stats; brackets, spaces and `@` in metric names are awkward for checkpoint-filename templates and monitor configs — consider matching the family convention. Also note this drops mAP@75/small/medium/large and AR that COCOeval provides for free; if the old evaluator surfaced per-class values, call out the reporting regression in the PR description.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
```diff
@@ -369,16 +368,25 @@ def on_validation_epoch_end(self):
         """
         self.status_logging_dict = {}
         if self.experiment_spec.dataset.val_data_sources.data_type == "OD":
-            eval_results = self.val_evaluator.summarize()
-            if self.trainer.is_global_zero and eval_results:
-                # Log main metrics (bbox/segm)
+            self.val_evaluator.synchronize_between_processes()
```
vpraveen-nv, line 395: Edge case the removed code explicitly handled (see the old fast_dev_run comment): if `update()` was never called on this rank (empty val split, `limit_val_batches=0`), `CocoEvaluator.synchronize_between_processes` hits `np.concatenate([])` and raises — and under DDP an asymmetric raise before the all_gather in `merge()` can hang the other ranks. grounding_dino shares this latent issue, so not a regression you introduced, but since this PR hardens OD evaluation, a cheap symmetric guard (skip summarize + log NaN/0 when `not self.val_evaluator.img_ids`) would close it for both val and test paths.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
```diff
@@ -473,7 +479,7 @@ def test_step(self, batch, batch_idx):
         )
         if self.experiment_spec.dataset.test_data_sources.data_type == "OD":
             res = {target['image_id'].item(): output for target, output in zip(targets, filtered_res)}
```
vpraveen-nv: The test path feeds `filtered_res` (post `threshold_predictions` — confidence + NMS filtering) into the COCO evaluator, while the validation path feeds raw `results`. Standard COCO mAP is computed over all ranked detections; confidence-thresholding beforehand systematically depresses AP and makes `test` mAP non-comparable to `val` mAP from the same checkpoint. This asymmetry pre-dates the PR, but now that both paths report official COCO numbers the discrepancy becomes visible — intentional? If yes, add a code comment; if not, use `results` here like the val path.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
```diff
@@ -484,13 +490,25 @@ def on_test_epoch_end(self):
         """
         self.status_logging_dict = {}
         if self.experiment_spec.dataset.test_data_sources.data_type == "OD":
-            eval_results = self.test_evaluator.summarize()
-            if self.trainer.is_global_zero and eval_results:
+            self.test_evaluator.synchronize_between_processes()
+            self.test_evaluator.overall_accumulate()
+            self.test_evaluator.overall_summarize(is_print=True)
```
vpraveen-nv: `overall_summarize(is_print=True)` runs on every rank (it must, since it follows the collective sync), so a multi-GPU `evaluate` prints the full COCO summary table N times interleaved. grounding_dino uses `is_print=False` and logs the numbers once under the rank-zero guard — suggest the same here.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
ramanathan831: Addressed in 5f8791a2: COCO metrics now use the sibling val_mAP/val_mAP50 convention and retain mAP75, size buckets, and AR statistics.

tests/cv_unit_test/mask_grounding_dino/test_coco_evaluator_routing.py
ramanathan831: Addressed in 5f8791a2: the routing test is marked with pytest.mark.cv_unit.

release/mask_grounding_dino_coco_evaluator_overlay/build_overlay.py
ramanathan831: Addressed in 5f8791a2: the product overlay directory was removed from the PR, including its hardcoded path and legacy-base tooling.

tests/cv_unit_test/mask_grounding_dino/test_coco_evaluator_routing.py
ramanathan831: Fixed in 5f8791a2: removed chr obfuscation; the assertions now use readable metric expressions.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
ramanathan831: Fixed in 5f8791a2: empty evaluator ranks synchronize, and summary is skipped when the global image set is empty.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
ramanathan831: Fixed in 5f8791a2: OD test evaluation now uses raw ranked results, matching validation.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
ramanathan831: Fixed in 5f8791a2: COCO overall_summarize now uses is_print=False on every rank.

nvidia_tao_pytorch/cv/mask_grounding_dino/model/pl_gdino_model.py
vpraveen-nv: Thanks for adding the guard — one placement issue: the crash happens *inside* `synchronize_between_processes()`, before the `all_gather` in `merge()`. On a rank where `update()` was never called, `self.eval_imgs[iou_type]` is `[]` and `np.concatenate` raises, so an empty rank still exits asymmetrically and the peers can hang.

Since `accumulate`/`summarize` are rank-local, the guard just needs to move ahead of the sync — e.g. have empty ranks append a correctly shaped empty array before syncing, or gate the whole OD block on an all-reduced global count so every rank reaches the `all_gather`. (Small note: `img_ids` is rank-local, so the check reflects the local shard rather than the global set.)

vpraveen-nv (approved):
approving.

PR #69, opened 2026-08-05
[TAO-0000][Test] Add DINO AutoML regression coverage

ramanathan831:
## Summary
- port DINO AutoML coverage from the discontinued GitLab MR
- preserve the current GitHub implementation while adding regression tests

## Root cause
The AutoML tests were never migrated from GitLab.

## Validation
- branch rebased onto current GitHub `main`
- `git diff --check`


ramanathan831:
Closing as a duplicate migration. The retained PR is #77, which contains the preferred/current-main version of this change. The source branch is being kept intact.

PR #71, opened 2026-08-05
[TAO-0000][Feature] Cache PyTorch CUDA builds

ramanathan831:
## Summary
- add reusable CUDA/PyTorch build-cache management
- integrate opt-in `--use-cache` behavior with the current deploy script
- avoid the obsolete hard-coded container digest from the old GitLab patch

## Root cause
The original GitLab source branch was deleted, so its final MR diff was ported semantically onto current GitHub `main`.

## Validation
- `bash -n release/docker/deploy.sh scripts/cuda_cache.sh`
- `git diff --check`


scripts/cuda_cache.sh
```diff
@@ -0,0 +1,81 @@
...
+
+CACHE_DIR="${HOME}/.tao_pytorch_cuda_cache"
+PROJECT_ROOT="${NV_TAO_PYTORCH_TOP:-$(pwd)}"
+BUILD_DIR="${PROJECT_ROOT}/build"
+HASH_FILE="${CACHE_DIR}/source.hash"
+
+generate_hash() {
+    {
+        find "$PROJECT_ROOT/nvidia_tao_pytorch" \
+            \( -name "*.cu" -o -name "*.cpp" \) -print0 2>/dev/null \
```
vpraveen-nv: **Cache-key gap:** headers aren't hashed. The CUDA extensions include `*.h`/`*.cuh` (e.g. MSDeformAttn's `cuda/` headers) — editing only a header leaves the hash unchanged, so `check` reports VALID and a stale `.so` gets packaged into the wheel. Add `-o -name "*.h" -o -name "*.cuh" -o -name "*.cc"` to the find.

scripts/cuda_cache.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -eo pipefail
+
+CACHE_DIR="${HOME}/.tao_pytorch_cuda_cache"
+PROJECT_ROOT="${NV_TAO_PYTORCH_TOP:-$(pwd)}"
+BUILD_DIR="${PROJECT_ROOT}/build"
+HASH_FILE="${CACHE_DIR}/source.hash"
```
vpraveen-nv, line 10: **Cache-key gap #2:** the hash covers sources only, not the toolchain. After a base-image bump (new PyTorch / CUDA / `TORCH_CUDA_ARCH_LIST`), the cache still validates and ABI-incompatible artifacts from the old toolchain get restored into a release build. Fold toolchain identity into `generate_hash` — e.g. the image digest from `docker/manifest.json` plus the effective `TORCH_CUDA_ARCH_LIST` — so an image bump invalidates the cache automatically.

scripts/cuda_cache.sh
```diff
@@ -0,0 +1,81 @@
...
+}
+
+case "${1:-}" in
+    restore)
+        if ! is_valid; then
+            echo "No valid cache found"
+            exit 1
+        fi
+        [ ! -d "$BUILD_DIR" ] || chmod -R u+w "$BUILD_DIR" 2>/dev/null || true
+        rm -rf "$BUILD_DIR"
```
vpraveen-nv, line 53: **Ownership hazard:** `build/` is produced inside the `tao_pt` container (typically as root), but `restore`/`save` run on the host as the regular user. `chmod -R u+w` as a non-owner fails, and `rm -rf` of root-owned files then fails too — with `set -eo pipefail` restore aborts, potentially after partially deleting `build/`. Either run the cache ops inside the container, or bail out early with a clear message when `build/` isn't owned by the invoking user.

release/docker/deploy.sh
```diff
@@ -90,10 +95,17 @@ if [ $BUILD_DOCKER = "1" ]; then
...
         echo "Building source code wheel ..."
-        # --no-tty so the wheel build works in non-interactive/CI contexts. Without it,
-        # tao_pt runs `docker run -it`, which aborts with "the input device is not a TTY".
-        # tao_pt --env 'TORCH_CUDA_ARCH_LIST="5.3 6.0 6.1 7.0 7.5 8.0 8.6 9.0"' -- bash /tao-pt/release/docker/build_wheel.sh
-        tao_pt --no-tty -- python setup.py bdist_wheel
+        if [ "$USE_CUDA_CACHE" = "1" ] && "$NV_TAO_PYTORCH_TOP/scripts/cuda_cache.sh" check; then
+            echo "Restoring cached CUDA build artifacts ..."
+            "$NV_TAO_PYTORCH_TOP/scripts/cuda_cache.sh" restore
+            tao_pt --no-tty -- python setup.py build_py
+            tao_pt --no-tty -- python setup.py bdist_wheel --skip-build
```
vpraveen-nv: If the restored cache is subtly stale/corrupt, this path fails at the wheel-existence check below with no automatic recovery. Since `--use-cache` is opt-in for CI speed, consider falling back to a full `bdist_wheel` (and re-`save`) when the cached path doesn't produce a wheel, so a bad cache degrades to a slow build instead of a failed deploy.

vpraveen-nv (commented):
not sure if this MR is a blocker, but thanks for putting this together.

scripts/cuda_cache.sh
ramanathan831: Addressed in 6fa1a509: CUDA cache hashing now includes .h, .cuh, and .cc files.

scripts/cuda_cache.sh
ramanathan831: Addressed in 6fa1a509: cache restore/save/clean check writability and fail with an ownership-specific message before chmod or deletion.

scripts/cuda_cache.sh
ramanathan831: Addressed in 6fa1a509: the cache key includes docker/manifest.json, the Dockerfile TORCH_CUDA_ARCH_LIST, and the effective environment value.

release/docker/deploy.sh
ramanathan831: Addressed in 6fa1a509: a failed cache restore or cached wheel build now falls back to a full bdist_wheel and re-saves the cache.

vpraveen-nv (approved):
All comments addressed — header + toolchain identity folded into the cache key, writability guard before destructive ops, and full-rebuild fallback in deploy.sh. LGTM.

PR #72, opened 2026-08-05
[TAO-0000][Bugfix] Port StyleGAN XL and NVDINOv2 fixes

ramanathan831:
## Summary
- port the still-relevant StyleGAN XL KPI and NVDINOv2 path fixes
- omit changes already present or superseded in current GitHub `main`

## Root cause
The original GitLab source branch was deleted; the final MR diff was reconciled against the current implementation.

## Validation
- Python compile checks for changed modules
- `git diff --check`


nvidia_tao_pytorch/ssl/nvdinov2/scripts/train.py
```diff
@@ -18,6 +18,13 @@
 
 def run_experiment(experiment_config, key):
     """Start the training."""
+    if (
+        experiment_config.get("train", {}).get("resume_training_checkpoint_path") ==
+        ""
+    ):
+        experiment_config.train.resume_training_checkpoint_path = None
+    if experiment_config.get("train", {}).get("pretrained_model_path") == "":
```
vpraveen-nv: Both normalizations collapse to a falsy check that also reads more naturally and avoids the awkward line wrap above:
```python
for key in ("resume_training_checkpoint_path", "pretrained_model_path"):
    if not experiment_config.train.get(key):
        experiment_config.train[key] = None
```
Also worth asking: other model families do this `"" -> None` normalization inside `initialize_train_experiment` / config validation rather than per-script — if more networks hit the same issue, the shared helper is the better home.

nvidia_tao_pytorch/sdg/stylegan_xl/model/sx_pl_model.py
```diff
@@ -487,8 +487,6 @@ def on_validation_epoch_end(self):
         # Log to status.json
         self.status_logging_dict = {}
         self.status_logging_dict["fid50k_full"] = float(fid_score)
```
vpraveen-nv, line 492: Question: what breaks with `seen_images`/`epoch` in the KPI dict — is this because status.json KPI consumers (AutoML monitor / TAO API) expect only optimizable metrics? Worth one sentence in the commit message or a code comment so the fields don't get re-added later by someone "improving" logging.

nvidia_tao_pytorch/ssl/nvdinov2/scripts/train.py
ramanathan831: Addressed in 505c0389: both fields use one falsy-value normalization loop. A shared-helper refactor is not needed for this single caller and can be centralized if additional networks require it.

nvidia_tao_pytorch/sdg/stylegan_xl/model/sx_pl_model.py
ramanathan831: Addressed in 505c0389: the code comment explains that seen_images and epoch are internal progress state, while status.json is reserved for numeric optimizable KPIs consumed by TAO and AutoML.

vpraveen-nv (approved):
Both comments addressed — normalization loop reads well and the KPI rationale comment is in place. LGTM.

PR #73, opened 2026-08-05
[TAO-0000][Bugfix] Route Mask2Former instance evaluation to mask AP

ramanathan831:
## Summary

Ports the completed Mask2Former product fix onto GitHub `main`:

- routes instance-segmentation evaluation to COCO mask AP;
- preserves semantic and panoptic metric routing;
- adds task-aware evaluation utilities and focused regression tests.

The exact legacy branch is preserved at `dev/ram/archive/rarunachalam/mask2former-instance-ap-product-fix`.

## Validation

- `6 passed` in `tests/cv_unit_test/mask2former/test_task_metric_routing.py`
- `git diff --check` passed.

## Migration

- GitHub base: `main`
- Legacy tip: `c2e86fe1`
- Main-native commit: `3e6f01fa`


nvidia_tao_pytorch/cv/mask2former/dataloader/pl_data_module.py
```diff
@@ -160,7 +149,21 @@ def test_dataloader(self):
         Returns:
             PyTorch DataLoader used for evaluation.
         """
-        return self.val_dataloader()
+        dataset_test = self._build_dataset("test", is_training=False)
```
vpraveen-nv: **Breaking change for `mask2former evaluate`.** Previously `test_dataloader()` returned `val_dataloader()`, so every existing eval spec configures `dataset.val`. Now evaluation reads `dataset.test`, whose `type` defaults to `ade` with an empty `annot_file` — the shipped specs (`experiment_specs/spec_coco.yaml` `test:` has only `img_dir`; same for `spec.yaml`) will crash or build the wrong dataset. Note `dataset.test` is also what `predict_dataloader` uses for label-free inference, so the same config node now serves two incompatible purposes (with vs. without annotations).

At minimum: update the shipped specs and docs, and consider falling back to `dataset.val` (with a deprecation warning) when `dataset.test` is unpopulated, so existing user specs and notebooks don't silently break.

nvidia_tao_pytorch/cv/mask2former/model/pl_model.py
```diff
@@ -394,98 +403,306 @@ def validation_step(self, batch, batch_idx):
...
+                name,
+                value,
+                on_step=False,
+                on_epoch=True,
+                prog_bar=True,
+                sync_dist=True,
+            )
+        if split == "val":
+            self.log("current_epoch", self.current_epoch, sync_dist=True)
+        return {name: str(value) for name, value in metrics.items()}
```
vpraveen-nv: KPI values are stringified here, but `status_logging_dict` is written to status.json as KPIs and consumed programmatically (TAO API / AutoML rank experiments by numeric KPI values). The semantic path returns floats (`miou`, `all_acc`). Return `float(value)` for consistency — a string `"0.42"` will break or silently mis-sort numeric consumers.

nvidia_tao_pytorch/cv/mask2former/utils/task_metrics.py
```diff
@@ -0,0 +1,133 @@
...
+            f"Unsupported evaluation split {split!r}; expected one of "
+            f"{sorted(SUPPORTED_EVALUATION_SPLITS)}."
+        )
+    return normalized
+
+
+def instance_metric_names(split):
+    """Return the stable COCO mask-AP metric names for an evaluation split."""
+    split = normalize_evaluation_split(split)
+    return f"segm_{split}_mAP", f"segm_{split}_mAP50"
```
vpraveen-nv: Two metric-name contract questions:

1. `segm_val_mAP` diverges from the detection-family convention — DINO/D-DETR log `val_mAP`/`val_mAP50`, which is what AutoML monitor configs and docs are written against. Was the `segm_` prefix deliberate?
2. `semantic_metric_names` renames panoptic-mode keys from `mIoU`/`all_acc` to `panoptic_{split}_semantic_mIoU_diagnostic` — any user or AutoML config monitoring `mIoU` on a panoptic run breaks, and tao-deploy's mask2former evaluation still reports `mIoU`, so pytorch/deploy metric parity diverges.

Both renames may be intentional, but they need explicit sign-off from TAO API / AutoML / docs owners and a release-note entry.

nvidia_tao_pytorch/cv/mask2former/dataloader/pl_data_module.py
```diff
@@ -110,32 +124,7 @@ def val_dataloader(self):
...
-                self.data_cfg.val.img_dir,
-                self.data_cfg.val.panoptic_dir,
-                cfg=self.data_cfg,
-                is_training=False,
-            )
-        else:
-            raise NotImplementedError(
-                f"The dataset type ({self.data_cfg.val.type}) is not supported."
-            )
+        dataset_val = self._build_dataset("val", is_training=False)
```
vpraveen-nv, line 127: Multi-GPU semantic metrics are still slightly biased: with `drop_last=False`, `DistributedSampler` pads ranks by *duplicating* samples, and those duplicates are double-counted in the semantic area sums that `_finalize_semantic_metrics` all-reduces (the instance path is safe — `CocoEvaluator` dedups image IDs via `np.unique`). Single-GPU is now exact, which is an improvement over the old `drop_last=True`, but consider per-rank dedup of the padded tail (or Lightning's `UnrepeatedDistributedSampler` semantics) if exact multi-GPU mIoU matters.

tests/cv_unit_test/mask2former/test_task_metric_routing.py
```diff
@@ -0,0 +1,234 @@
...
+        finalize_instance_evaluator(evaluator, split="test")
+
+    evaluator = _FakeCocoEvaluator()
+    evaluator.coco_eval = {}
+    with pytest.raises(RuntimeError, match="did not produce segm"):
+        finalize_instance_evaluator(evaluator, split="test")
+
+
+@pytest.mark.cv_unit
+def test_runtime_sources_preserve_the_task_and_split_contracts():
```
vpraveen-nv: This test asserts raw substrings of `pl_model.py`/`datasets.py`/`pl_data_module.py` source text (e.g. `'if self.mode == "instance"'`, `datamodule_source.count("drop_last=False") >= 3`). It's a change-detector: any legitimate refactor fails it, while a real behavioral regression that keeps the strings passes. Consider dropping it in favor of the behavioral tests above, or exercising the actual code paths with a stubbed trainer instead of grepping source.

vpraveen-nv (commented):
minro comments

nvidia_tao_pytorch/cv/mask2former/dataloader/pl_data_module.py
ramanathan831: Addressed in 499d6474: validation and test use Lightnings UnrepeatedDistributedSampler, preventing padded duplicate samples from biasing globally reduced semantic metrics.

nvidia_tao_pytorch/cv/mask2former/utils/task_metrics.py
ramanathan831: Addressed in 499d6474: instance metrics use the stable val_mAP and val_mAP50 family names, and panoptic semantic diagnostics retain mIoU and all_acc for monitor and tao-deploy parity.

nvidia_tao_pytorch/cv/mask2former/dataloader/pl_data_module.py
ramanathan831: Addressed in bc24d080: test_dataloader falls back to the annotated val split with a warning when test has no annotations, preserving existing evaluation specs while leaving prediction label-free.

nvidia_tao_pytorch/cv/mask2former/model/pl_model.py
ramanathan831: Addressed in bc24d080: status KPI values are returned as floats rather than strings.

tests/cv_unit_test/mask2former/test_task_metric_routing.py
ramanathan831: Addressed in 71c77916: the brittle raw-source integration test was removed; routing and evaluator behavior remain covered by the data-only and fake-evaluator tests.

nvidia_tao_pytorch/cv/mask2former/model/pl_model.py
```diff
@@ -394,98 +403,303 @@ def validation_step(self, batch, batch_idx):
...
+        if self.mode == "instance":
+            self.test_coco_evaluator = self._create_instance_evaluator("test")
 
     def test_step(self, batch, batch_idx):
-        """Test step"""
-        return self.validation_step(batch, batch_idx)
+        """Evaluate a standalone test batch using the configured task metric."""
+        inputs = batch['images']
+        if inputs.shape[0] > 1:
+            assert len(self.cfg.dataset.test.target_size) > 0, \
```
vpraveen-nv: Small interaction with the new val fallback: when `dataset.test` has no annotations, the test loader takes `batch_size` from `dataset.val` and the dataset resizes with `val.target_size`, but this assert still checks `dataset.test.target_size`. A user relying on the fallback with `val.batch_size > 1` hits a spurious assertion, and setting only `test.target_size` would silence it without changing the actual resize (which uses val). Consider asserting on the `target_size` of the split actually in use.

vpraveen-nv (approved):
All comments addressed — verified the val fallback with warning, numeric KPIs, the metric-name reverts, and the UnrepeatedDistributedSampler change in the updated diff. One minor note left inline on the fallback/assert interaction, no need to block on it. LGTM.

PR #74, opened 2026-08-05
[TAO-0000][Bugfix] Correct OneFormer PTM, metrics, and runtime overlay

ramanathan831:
## Summary

Ports the completed OneFormer runtime fixes onto GitHub `main`:

- correct PTM loading and task-specific distributed metric reduction;
- add panoptic-quality handling and explicit metric routing;
- add the pinned SQSH source-overlay tooling with separate base/source audits;
- add focused product and overlay regression coverage.

The exact legacy branch is preserved at `dev/ram/archive/rarunachalam/oneformer-runtime-product-fixes`.

## Validation

- `16 passed` across the focused OneFormer product-fix and runtime-overlay suites.
- `git diff --check` passed.

## Migration

- GitHub base: `main`
- Legacy tip: `1752ec2c`
- Main-native commits: `bfeee11b`, `718a5bf3`, `10262648`


nvidia_tao_pytorch/cv/oneformer/model/pl_oneformer.py
```diff
@@ -621,86 +735,165 @@ def validation_step(self, batch, batch_idx, dataloader_idx=0):
...
-        """Compute IoU metrics per dataloader and combined, then log them.
+        """Globally reduce sufficient statistics, then compute task metrics.
 
         Args:
             outputs_by_dl: dict mapping dataloader_idx -> list of metric dicts.
             dataset_names: list of human-readable names (or None for single dataset).
             prefix: optional prefix for logged metric keys (e.g. "test_").
         """
+        if not outputs_by_dl:
+            raise RuntimeError("OneFormer evaluation produced no metric observations.")
```
vpraveen-nv: **Potential DDP hang instead of clean failure.** `_global_semantic_stats`/`_global_panoptic_stats` call `distributed_sum_array` (an `all_reduce` collective). If one rank ends up with an empty `outputs_by_dl` (tiny val set with more GPUs than shards, `limit_val_batches=0`, or a dataloader index present on some ranks only), that rank raises here while the others block forever inside `all_reduce` — the job hangs until NCCL timeout. Consider making the collective unconditional (reduce a zero array even when a rank has no local outputs, iterating over *expected* dataloader indices from the datamodule rather than observed `outputs_by_dl` keys), and raise only after reduction if the global sums are all zero.

release/oneformer_runtime_overlay/build_overlay.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def _base_blob(repo_root, path):
+    result = subprocess.run(
+        ("git", "show", f"{BASE_COMMIT}:{path}"),
+        cwd=repo_root,
+        check=False,
+        capture_output=True,
+    )
+    return result.stdout if result.returncode == 0 else None
```
vpraveen-nv, line 56: **`_base_blob` conflates "commit missing" with "path absent at base commit".** `git show {BASE_COMMIT}:{path}` fails identically when `99741bc8` isn't in the local clone — the *likely* case, since GitHub `main` is a squashed migration root. Then every payload file silently gets `base_sha256: null`, and `install_overlay.py` later fails with the misleading "A new overlay target already exists in the pinned base" for the files that do exist in the SQSH. Verify the commit up front (`git cat-file -e {BASE_COMMIT}^{{commit}}`) and only treat a path-level miss as "new file".

release/oneformer_runtime_overlay/build_overlay.py
```diff
@@ -0,0 +1,179 @@
...
+
+BASE_REF = "origin/release/7.1.0"
+BASE_COMMIT = "99741bc8229617d0d3dd52e30540111d55efd1af"
+CONTAINER = {
+    "image_reference": (
+        "nvcr.io/nvstaging/tao/tao-toolkit-pyt:"
+        "7.1.0-rc-245-multiarch"
+    ),
+    "sqsh_path": (
+        "/lustre/fsw/portfolios/edgeai/users/rarunachalam/"
```
vpraveen-nv: **Personal, environment-specific paths hardcoded in product-repo tooling.** The pinned SQSH path, the README's `/localhome/local-rarunachalam/...` build location, and the Lustre artifact destinations are one user's cluster layout, and the pinned image is an RC (`7.1.0-rc-245`) that will age out. This conflicts with the repo convention against hardcoded paths. Does this overlay belong on product `main` at all (vs. an internal ops repo or evidence linked from the PR)? If it stays, make the SQSH path/output locations CLI arguments with the pinned sha256 as the only committed invariant.

nvidia_tao_pytorch/config/oneformer/evaluate.py
```diff
@@ -19,6 +19,17 @@
 class OneFormerEvaluateConfig(EvaluateConfig):
     """Evaluation configuration for OneFormer."""
 
+    task: str = STR_FIELD(
```
vpraveen-nv, line 22: **Cross-project schema drift with tao-deploy.** `nvidia_tao_deploy/config/oneformer/evaluate.py` has no `task` field in its `OneFormerEvaluateConfig`. Users routinely reuse the same experiment spec across `tao_pt` and `tao_deploy`, and all nine shipped specs in this PR now carry `evaluate.task: semantic` — a structured-config merge on the deploy side will reject the unknown key. Add a matching (even if ignored) `task` field to the tao-deploy schema in a companion PR, or confirm the deploy entrypoint tolerates unknown evaluate keys.

nvidia_tao_pytorch/cv/oneformer/dataloader/datasets.py
```diff
@@ -140,9 +146,12 @@ def __init__(  # pylint: disable=too-many-arguments
             self.instance_prob = cfg.dataset.task_prob_train.instance
             self.panoptic_prob = cfg.dataset.task_prob_train.panoptic
         else:
-            self.semantic_prob = cfg.dataset.task_prob_val.semantic
-            self.instance_prob = cfg.dataset.task_prob_val.instance
-            self.panoptic_prob = cfg.dataset.task_prob_val.panoptic
+            self.evaluation_task = str(cfg.evaluate.task).lower()
```
vpraveen-nv, line 149: `dataset.task_prob_val` (`DICT_FIELD` in `config/oneformer/dataset.py`) is now dead config: the non-training branch no longer reads it, but the field remains in the schema and existing user specs, silently ignored. Remove/deprecate it (with a note in the field description) or warn when it's set to a non-default value. Also, this validation duplicates the identical check in `pl_oneformer.__init__` — the config field already declares `valid_options="semantic,panoptic"`, so one authoritative place would be cleaner.

nvidia_tao_pytorch/cv/oneformer/utils/panoptic_quality.py
```diff
@@ -0,0 +1,278 @@
...
+            "Prediction ID map contains segments missing from segments_info: "
+            f"{sorted(unknown_pred)}."
+        )
+    if unknown_gt:
+        raise ValueError(
+            "Ground-truth ID map contains segments missing from segments_info: "
+            f"{sorted(unknown_gt)}."
+        )
+
+    # Resizing can remove a tiny GT segment completely. Such a segment is not
```
vpraveen-nv, line 103: **Protocol deviation from official panopticapi worth documenting loudly.** Two differences from COCO PQ: (1) GT segments that vanish after resize are dropped instead of counted as FN (official evaluation runs at original PNG resolution, so this never happens there — dropping them inflates PQ); (2) PQ is computed on the resized+padded `target_size` raster, not original resolution. The README/config text says "COCO-style PQ", which readers may take as comparable to published numbers. State explicitly that values are computed at network input resolution and aren't directly comparable to official panopticapi results, and consider logging the count of dropped GT segments so the inflation is visible.

tests/cv_unit_test/oneformer/test_runtime_product_fixes.py
```diff
@@ -0,0 +1,285 @@
...
+    distributed = _FakeDistributed
+
+    @staticmethod
+    def as_tensor(value, dtype=None, device=None):
+        del dtype, device
+        return _FakeTensor(value)
+
+
+_real_torch = sys.modules.get("torch")
+sys.modules["torch"] = _FakeTorch
```
vpraveen-nv: Swapping `sys.modules["torch"]` at module import time mutates global interpreter state outside any fixture, breaks under `pytest-xdist`, and leaves `metric_utils` permanently bound to `_FakeTorch` for the rest of the session. Prefer a `monkeypatch`-based fixture, or import the real module and patch `torch.distributed.is_initialized`/`all_reduce`. Relatedly, asserting on `ast.unparse` of production source (e.g. `"sync_dist=False" in ...`) pins implementation details — it fails on harmless refactors while not catching real behavioral regressions; a small LightningModule-level test with a stubbed trainer would guard the actual contract.

vpraveen-nv (commented):
minor comments. deploy sync is required though.

release/oneformer_runtime_overlay/build_overlay.py
ramanathan831: Addressed in fd53feea: personal SQSH, checkout, and Lustre paths were removed from product tooling and README examples; deployment paths are inputs and the committed container invariant is the image digest.

nvidia_tao_pytorch/cv/oneformer/dataloader/datasets.py
ramanathan831: The validation branch still consumes task_prob_val in COCOUnifiedDataset for unified task sampling; it is not silently ignored. The schema description now documents that backward-compatible use.

nvidia_tao_pytorch/cv/oneformer/utils/panoptic_quality.py
ramanathan831: Addressed in 6b44a736: the OneFormer README now states the resized/padded input-resolution contract, dropped-GT behavior, and that the values are not directly comparable with official panopticapi results.

release/oneformer_runtime_overlay/build_overlay.py
ramanathan831: Addressed in 6b44a736: the pinned base commit is verified with git cat-file before path-level git show failures are interpreted as new files.

nvidia_tao_pytorch/cv/oneformer/model/pl_oneformer.py
ramanathan831: Addressed in 6b44a736: evaluation now iterates over expected dataloader indices and supplies zero local statistics for empty ranks, so every rank enters the reduction collective.

tests/cv_unit_test/oneformer/test_runtime_product_fixes.py
ramanathan831: Addressed in 6b44a736: the fake torch module is scoped with mock.patch.dict and restored immediately; the remaining source contract checks are data-only and do not mutate interpreter state.

nvidia_tao_pytorch/config/oneformer/evaluate.py
ramanathan831: This PR stays scoped to tao-pytorch. The tao-deploy schema field requires a separate companion change and deploy-owner approval; no unsafe cross-repository edit was made here.

vpraveen-nv: Agreed a separate change is the right place for the deploy schema. One concern ships with this PR though: the specs here now carry `evaluate.task`, and tao-deploy's `OneFormerEvaluateConfig` doesn't have that field yet, so a shared spec would fail the deploy-side config merge. Could we link a tracking issue/PR for the deploy schema, or confirm the deploy entrypoint tolerates unknown `evaluate` keys?

nvidia_tao_pytorch/cv/oneformer/dataloader/datasets.py
vpraveen-nv: I may be misreading the latest revision, but it looks like the non-training reads of `task_prob_val` are now removed (the val path uses `cfg.evaluate.task`), and `config/oneformer/dataset.py` isn't part of the PR, so the field appears unused outside training with its description unchanged. Could you double-check? A short deprecation note in the field description — or a warning when it's set to a non-default value — would fully cover this.

nvidia_tao_pytorch/cv/oneformer/model/pl_oneformer.py
```diff
@@ -621,86 +735,173 @@ def validation_step(self, batch, batch_idx, dataloader_idx=0):
...
-
+        # Lightning can invoke this hook on a rank that received no samples.
+        # Still enter every configured dataloader's collective with zero local
+        # statistics so that ranks cannot diverge or deadlock.
+        if dataset_names is not None:
+            expected_indices = range(len(dataset_names))
+        else:
+            expected_indices = (0,)
+        outputs_by_dl = {
+            index: outputs_by_dl.get(index, []) for index in expected_indices
```
vpraveen-nv, line 753: Minor follow-up on the DDP fix: with the pre-collective raise removed, a fully empty evaluation now completes and writes all-zero KPIs instead of failing. Since the reduction now runs symmetrically, a post-reduction check (raise or error-log when the global sums are all zero) would restore the failure signal while staying deadlock-safe.

vpraveen-nv (approved):
Approved pending these fixes.

PR #75, opened 2026-08-05
[TAO-0000][Bugfix] Correct SegFormer PTM loading and distributed metrics

ramanathan831:
## Summary

Ports the completed SegFormer PTM/runtime fixes onto GitHub `main`:

- load trainable PTMs as pretrained weights instead of trainer resume state;
- support prefixed backbone checkpoint keys;
- aggregate validation metrics correctly across distributed ranks;
- add focused PTM-loading and distributed-metric regression coverage.

The exact legacy branch is preserved at `dev/ram/archive/rarunachalam/segformer-ptm-loader-product-fix-v2`.

## Validation

- Python compilation passed for all changed production and test modules.
- Focused pytest collection is environment-blocked because the available migration venv does not include `torch`; this is not a test assertion failure.
- `git diff --check` passed.

## Migration

- GitHub base: `main`
- Legacy tip: `2681dea4`
- Main-native commits: `693bf8f4`, `5853e12f`, `483d0255`


nvidia_tao_pytorch/cv/segformer/utils/checkpoint.py
```diff
@@ -0,0 +1,299 @@
...
+    # are removed so similarly named components inside a backbone are intact.
+    while candidate.startswith(("module.", "_orig_mod.")):
+        prefix = "module." if candidate.startswith("module.") else "_orig_mod."
+        candidate = candidate[len(prefix):]
+        add(candidate)
+
+    # A Lightning state_dict contains ``model.*`` keys, while this helper
+    # loads the wrapped torch model whose state_dict starts at ``backbone.*``
+    # and ``decoder.*``. Repeated wrappers cover nested Lightning modules.
+    while candidate.startswith("model."):
```
vpraveen-nv: `_candidate_model_keys` strips `module.`/`_orig_mod.` first, then `model.` — but the loops don't alternate, so an interleaved key like `model._orig_mod.backbone.weight` (Lightning module wrapping a `torch.compile`d model) never normalizes to `backbone.weight`: after stripping `model.` the remaining `_orig_mod.` prefix is never revisited. `_candidate_backbone_keys` below already handles this correctly with a single loop over all three prefixes — consider the same pattern here:
```python
while True:
    prefix = next((p for p in ("module.", "_orig_mod.", "model.") if candidate.startswith(p)), None)
    if prefix is None:
        break
    candidate = candidate[len(prefix):]
    add(candidate)
```

nvidia_tao_pytorch/cv/segformer/utils/checkpoint.py
```diff
@@ -0,0 +1,299 @@
...
+        len(report.loaded_keys),
+        component,
+        label,
+        list(report.loaded_keys),
+    )
+    if report.missing_keys:
+        logging.warning(
+            "SegFormer tensors kept at initialization (%d): %s",
+            len(report.missing_keys),
+            list(report.missing_keys),
```
vpraveen-nv: `_log_report` dumps the *full* sorted key lists — `loaded_keys` at INFO plus `missing_keys` at WARNING. For a real FAN/ViT-adapter SegFormer that's hundreds of lines per PTM load, and for a backbone-only PTM, `missing_keys` will contain every decoder tensor and BN buffer — a misleading wall of warnings for a normal load. Suggest counts + truncated sample at INFO/WARNING, full lists at DEBUG. Related: if nothing downstream parses the `SEGFORMER_PRETRAINED_LOAD_REPORT` sha256 "receipt", consider dropping it — it reads like one-off qualification evidence embedded in product code.

nvidia_tao_pytorch/cv/segformer/utils/iou_metric.py
```diff
@@ -150,8 +153,43 @@ def update_cm(self, pr: torch.Tensor, gt: torch.Tensor):
             )
             self.update(area_intersect, area_union, area_pred_label, area_label)
 
-    def get_scores(self):
-        """get scores from confusion matrix"""
+    def _globally_reduce_sufficient_statistics(self, device=None):
+        """Sum metric sufficient statistics across all distributed ranks."""
+        statistics = np.stack(
```
vpraveen-nv: If the meter was never fed (`initialize` not called — e.g. a rank with zero val batches, or an empty split), `self.area_intersect` etc. are `None` and `np.stack` raises. Pre-existing code also crashed, but now the DDP failure mode is worse: the raising rank exits while peers block in `dist.all_reduce` — a hang instead of a clean error. Consider falling back to `np.zeros((4, self.n_class))` when `not self.initialized`, so the collective always completes and empty ranks contribute nothing (also numerically correct for a SUM reduction).

nvidia_tao_pytorch/cv/segformer/model/segformer_pl_model.py
```diff
@@ -333,7 +334,7 @@ def _clear_cache(self):
 
     def _collect_epoch_states(self):
         """Collect evaluation metrics for each epoch (train, val, test)"""
-        scores, mean_score_dict = self.running_metric.get_scores()
+        scores, mean_score_dict = self.running_metric.get_scores(device=self.device)
```
vpraveen-nv, line 337: Now that `get_scores` returns globally-reduced, rank-identical values, `sync_dist=True` on the `log_dict` calls below is redundant — extra collectives per metric key just to average identical numbers. Consider `sync_dist=False` here (and on the `val_acc`/`val_miou` logs in `on_validation_epoch_end`).

nvidia_tao_pytorch/cv/segformer/model/segformer_pl_model.py
```diff
@@ -373,17 +374,25 @@ def training_step(self, batch, batch_idx):
         )
         return loss
 
+    def _write_status_kpis(self, kpis, message):
+        """Write status KPIs exactly once from the global rank-zero process."""
+        if get_global_rank() != 0:
```
vpraveen-nv, line 379: Minor: `StatusLogger.write` is already `@rank_zero_only` in `core/loggers/api_logging.py`, so nonzero ranks never wrote to status.json — the docstring's "exactly once from the global rank-zero process" and the test's "cannot race the global writer" slightly misstate the prior failure mode. The guard is fine as defense-in-depth (and keeps nonzero-rank `status_logger.kpi` state clean), but the real fix is that KPI *values* are now rank-identical. Worth tightening the wording so future readers don't hunt for a race that didn't exist.

tests/cv_unit_test/segformer/test_distributed_metrics.py
```diff
@@ -0,0 +1,131 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Pure mocked-distributed tests for SegFormer metric finalization."""
```
vpraveen-nv, line 4: `@pytest.mark.segformer` (used throughout this file and test_pretrained_weights.py) isn't registered in `pytest.ini` `markers`. Without `--strict-markers` it's only a `PytestUnknownMarkWarning` today, but please register it (like `ocdnet`/`ocrnet`) or drop it so marker-based selection stays reliable.

vpraveen-nv (commented):
minor comments.

nvidia_tao_pytorch/cv/segformer/utils/iou_metric.py
ramanathan831: Addressed in 3cb03caa: an uninitialized meter contributes zero sufficient statistics and still participates in all_reduce.

nvidia_tao_pytorch/cv/segformer/utils/checkpoint.py
ramanathan831: Addressed in 3cb03caa: INFO and WARNING logs now contain counts and short samples; complete key lists are available at DEBUG, while the deterministic load receipt remains for provenance.

nvidia_tao_pytorch/cv/segformer/utils/checkpoint.py
ramanathan831: Addressed in 3cb03caa: wrapper removal now loops until no supported leading module, _orig_mod, or model prefix remains, including interleaved wrappers.

nvidia_tao_pytorch/cv/segformer/model/segformer_pl_model.py
ramanathan831: Addressed in 3cb03caa: globally reduced values are logged with sync_dist=False, including val_acc and val_miou.

nvidia_tao_pytorch/cv/segformer/model/segformer_pl_model.py
ramanathan831: Addressed in 3cb03caa: the status KPI docstring now describes the rank-zero guard without implying a prior status.json race.

tests/cv_unit_test/segformer/test_distributed_metrics.py
ramanathan831: Addressed in a8537e21: the segformer pytest marker is now registered in pytest.ini.

nvidia_tao_pytorch/cv/segformer/model/segformer_pl_model.py
vpraveen-nv: The `_write_status_kpis` docstring is fixed — thanks. The other half of the comment still stands though: the test docstring in `tests/cv_unit_test/segformer/test_distributed_metrics.py` still reads "Nonzero ranks cannot race the global writer for status.json." — the exact race-that-didn't-exist wording. Since `StatusLogger.write` was already `@rank_zero_only`, suggest something like "Nonzero ranks leave KPI state untouched; rank zero remains the only writer." Purely cosmetic; feel free to fold into any next push.

vpraveen-nv (approved):
All comments addressed — verified the prefix-normalization loop, the zeros fallback in the metric all-reduce, the sync_dist cleanup, and the marker registration in the updated diff. One cosmetic docstring note left inline, no need to block on it. LGTM.

PR #76, opened 2026-08-05
[TAO-0000][Bugfix] Improve PyTorch entrypoint failure diagnostics

ramanathan831:
## Summary

Port the entrypoint completion, interruption, return-code, and exception diagnostics from the discontinued GitLab MR.

This is the GitHub-main replacement for work developed before the GitLab repositories were discontinued.

## Migration contract

- Base: `main`
- Source branch: `dev/ram/debug-pyt-main`
- Branch prefix: `dev/ram/`
- Repository history, not an image or generated runtime patch, is the source of truth.
- No credentials or user-specific runtime paths are included.


vpraveen-nv (approved):

nvidia_tao_pytorch/core/entrypoint.py
```diff
@@ -321,12 +321,19 @@ def launch(args, unknown_args, subtasks, network=None):
                 log_target.write(last_progress_bar_line + '\n')
                 log_target.flush()
             if proc.returncode == 0:
+                logging.info("Command completed successfully with return code 0")
                 process_passed = True
+            else:
+                logging.info(f"Command completed with return code {proc.returncode}")
 
     except (KeyboardInterrupt, SystemExit):
+        logging.info("Command was interrupted")
```
vpraveen-nv: This duplicates the `logging.exception("Command was interrupted")` on the next line, which already emits the message (at ERROR) plus the traceback. Drop this one, or you can make the other one info

nvidia_tao_pytorch/core/entrypoint.py
```diff
@@ -321,12 +321,19 @@ def launch(args, unknown_args, subtasks, network=None):
...
         logging.exception("Command was interrupted")
         process_passed = True
     except Exception as e:
+        logging.info("Command encountered an exception")
+        import traceback
```
vpraveen-nv: `logging.info("Command encountered an exception")` + `logging.info(traceback.format_exc())` is equivalent to a single `logging.exception("Command encountered an exception")` (or `logging.error(..., exc_info=True)`) — and at ERROR level the traceback isn't silently filtered when the root logger is set above INFO. If you keep `traceback`, move the import to module top per convention (pylint C0415 `import-outside-toplevel` may also flag this in the changed-files lint).

vpraveen-nv (commented):
minor comments.

nvidia_tao_pytorch/core/entrypoint.py
ramanathan831: Addressed in e19f8fad: the info plus manual traceback sequence was replaced with one logging.exception call.

nvidia_tao_pytorch/core/entrypoint.py
ramanathan831: Addressed in e19f8fad: the duplicate interruption log was removed, leaving the exception log with its traceback.

vpraveen-nv (approved):
Both comments addressed — duplicate log dropped and the manual traceback replaced with a single `logging.exception`. LGTM.

PR #77, opened 2026-08-05
[TAO-0000][Feature] Add DINO AutoML integration tests

ramanathan831:
## Summary

Port the DINO AutoML test harness and ASHA, Bayesian, Hyperband, and PBT coverage onto current GitHub main.

This is the GitHub-main replacement for work developed before the GitLab repositories were discontinued.

## Migration contract

- Base: `main`
- Source branch: `dev/ram/dino-automl-tests-main`
- Branch prefix: `dev/ram/`
- Repository history, not an image or generated runtime patch, is the source of truth.
- No credentials or user-specific runtime paths are included.


tests/cv_unit_test/dino/automl/harness.py
```diff
@@ -0,0 +1,345 @@
...
+    search_roots.extend(Path(path) for path in sys.path if path)
+
+    for search_root in search_roots:
+        candidate = search_root / DEFORMABLE_ATTN_OPS / lib_name
+        if candidate.exists() and not candidate.resolve().is_relative_to(REPO_ROOT):
+            return candidate
+    return None
+
+
+_patch_deformable_attention_loader_for_release_image()
```
vpraveen-nv: This permanently monkeypatches `deformable_ops_functions.load_ops` and `deformable_ops_modules.load_ops` at **import time** for the rest of the pytest session. CI runs the whole `tests/` tree in a single invocation (`ci/run_functional_tests.py` → `pytest <tests> --testmon`), so merely *collecting* this package changes the ops loader used by every unrelated deformable_detr/DINO test that runs afterwards. Move the patch into a package-scoped fixture that restores the originals on teardown.

pytest.ini
```diff
@@ -13,6 +13,9 @@ markers =
     infer: tests related to inference scripts (deselect with '-m "not infer"')
     infer_onnx: tests related to ONNX-based inference scripts (deselect with '-m "not infer_onnx"')
     cv_unit: marks cv unit test (deselect with '-m "not cv_unit"')
+    dino: marks DINO related tests (deselect with '-m "not dino"')
+    integration: marks integration tests (deselect with '-m "not integration"')
+    automl: marks AutoML integration tests (deselect with '-m "not automl"')
```
vpraveen-nv, line 18: Registering the markers is good, but nothing actually deselects them: functional CI runs by path with no `-m` filter, and these live under `tests/cv_unit_test/` despite being multi-training GPU integration tests. If `tao-automl` is ever importable in the CI image, the unit-test job silently starts running ~10 real DINO trainings — and even when skipped, the heavy imports in `harness.py` execute at collection. Suggest gating the package behind an opt-in env var (skip in `conftest.py` unless e.g. `TAO_AUTOML_TESTS=1`) or adding `-m "not automl"` to the CI invocation, and consider `tests/integration/` instead of `tests/cv_unit_test/`.

tests/cv_unit_test/dino/automl/harness.py
```diff
@@ -0,0 +1,345 @@
...
+            limit_val_batches=0,
+            callbacks=[epoch_trace],
+        )
+        trainer.fit(model, dm, ckpt_path=str(resume_checkpoint) if resume_checkpoint else None)
+
+        metric = _metric_from_trainer(trainer)
+        checkpoint_path = _latest_checkpoint(run_dir)
+        # Keep ranking deterministic for algorithm-contract assertions while
+        # still deriving the objective from a real training result.
+        reported_metric = metric + (float(rec.id) * 1e6)
```
vpraveen-nv: Since real losses are ≪ 1e6, the reported metric is a pure function of `rec.id` — the lowest id always wins. That makes the promotion/exploit assertions in the Hyperband/ASHA/PBT tests unfalsifiable for ranking bugs: a brain that promotes by insertion order (ignoring reported metrics entirely) passes identically. Invert the offset in at least one test (e.g. `-rec.id * 1e6`, so a *later* recommendation wins) to prove the algorithms rank by the reported metric rather than submission order.

tests/cv_unit_test/dino/automl/harness.py
```diff
@@ -0,0 +1,345 @@
...
+    candidates.append(REPO_ROOT.parent / "tao-automl" / "src")
+
+    for candidate in candidates:
+        if (candidate / "tao_automl").exists():
+            sys.path.insert(0, str(candidate))
+            return candidate
+    return None
+
+
+_add_tao_automl_source_path()
```
vpraveen-nv: `sys.path.insert(0, ...)` at import time with an environment-derived path leaks into the whole pytest session and takes precedence over installed packages for *all* subsequent tests. If both an installed `tao-automl` and a sibling checkout exist, the checkout silently wins. Prefer doing this in the package `conftest.py` (colocated with the skip logic), and consider appending rather than inserting at position 0 so an installed package wins by default.

tests/cv_unit_test/dino/automl/test_asha.py
```diff
@@ -0,0 +1,64 @@
...
+            "automl_max_concurrent": 2,
+            "automl_max_trials": 2,
+            "automl_min_top_configs": 1,
+        },
+    )
+
+    first_rung = runner.next_recommendations()
+    assert len(first_rung) == 2
+    assert [rec.early_stop_epoch for rec in first_rung] == [1, 1]
+    [dino_case.train(rec, target_epoch=rec.early_stop_epoch) for rec in first_rung]
```
vpraveen-nv: List comprehension used purely for side effects (also in `test_pbt.py`). The `static-tests` workflow runs pylint on changed files and this typically trips `W0106 expression-not-assigned`. Use a plain `for` loop.

tests/cv_unit_test/dino/automl/harness.py
```diff
@@ -0,0 +1,345 @@
...
+        self.brain.save_state()
+        return recs
+
+    def finish_if_ready(self):
+        self.brain.generate_recommendations(self.history)
+        self.brain.save_state()
+
+    def is_complete(self):
+        self.controller.history = self.history
+        self.controller._next_id = self._next_id
```
vpraveen-nv: Reaching into `Controller._next_id` (and reassigning `controller.history`) couples the harness to tao-automl private internals — a rename there breaks these tests opaquely, and tao-automl versions independently of tao-pytorch, so this is cross-repo fragility. Consider exposing a public `Controller.is_complete(history, next_id)`-style API in tao-automl and using that here.

tests/cv_unit_test/dino/automl/harness.py
ramanathan831: Addressed in 148423a1: source-path setup is colocated in conftest.py and appended, so installed tao-automl wins over a sibling checkout.

tests/cv_unit_test/dino/automl/harness.py
ramanathan831: Addressed in dc2652e9: the deformable-ops patch now lives in a package-scoped fixture and restores both original loaders during teardown.

tests/cv_unit_test/dino/automl/test_asha.py
ramanathan831: Addressed in 148423a1: side-effect-only list comprehensions were replaced with plain for loops.

tests/cv_unit_test/dino/automl/harness.py
ramanathan831: Addressed in 148423a1: the reported objective now subtracts rec.id*1e6, making the later recommendation preferable and testing ranking rather than insertion order.

pytest.ini
ramanathan831: Addressed in 148423a1: the package is opt-in via TAO_AUTOML_TESTS=1 and skips before heavy imports; the registered markers remain available for selection.

tests/cv_unit_test/dino/automl/harness.py
ramanathan831: Addressed in dc2652e9: the harness no longer touches Controller._next_id or controller.history; completion uses the brain public done() API when available, otherwise Controller.is_complete().

vpraveen-nv: Dropping the private `_next_id`/`history` pokes is the right direction — but the fallback breaks `test_bayesian`: `Bayesian` doesn't define `done()`, so `is_complete()` falls back to `Controller.is_complete()`, which counts completed recommendations in `controller.history`. The harness never registers its recommendations with the controller, so that history stays empty and the final `assert runner.is_complete()` can never pass.

Two options: drive recommendations through the controller's public path (`controller.next_recommendation()` + `controller.report_result()`), or use a completion API that accepts the externally tracked history.

PR #84, opened 2026-08-06
[TAO-2489][Feature] DINOv3 LoRA: key-preserving LoRALinear module and config

vpraveen-nv:
Adds `ssl/dinov3/model/lora.py` and the DINOv3 config blocks for LoRA + preservation
continual pre-training. No behaviour change on its own: `model.lora.enable` defaults to `false`,
and nothing is wired in until #85.

`LoRALinear` subclasses `nn.Linear` and **adopts the wrapped layer's existing `weight`/`bias`
Parameter objects**, so backbone state-dict keys stay exactly the stock keys plus
`lora_A`/`lora_B`. This is deliberately *not* a `.original`-nesting wrapper — four DINOv3
subsystems depend on stable backbone key names: `checkpoint_remap.timm_to_tao` /
`convert_ssl_to_timm`, `CustomModelCheckpoint`, `_sync_gram_teacher`, and
`restore_pretrained_weights`.

`lora_B` is zero-initialized, so a freshly injected backbone is an exact identity of the one it
replaced and step-0 SSL behaviour is unchanged.

Config: `LoRAConfig` gains `dropout`, `target_modules`, `num_last_blocks`; new
`PreservationConfig` (`enable`, `cls_mse_weight`, `cls_cosine_weight`). `dinov3 default_specs`
emits both blocks.

JIRA: TAO-2489.

### Test evidence

`pytest tests/ssl_unit_test/dinov3`, measured against `main`:

| | result |
|---|---|
| baseline (`main`) | 75 passed, 4 skipped |
| with the full stack | **100 passed, 4 skipped** |

All 25 tests are additions; the same 4 skip for the same reason, so there are no regressions.
`pylint --rcfile .pylintrc` 10.00/10 on the changed sources; `pydocstyle` and `flake8` clean.

**Gate G1 (static invariants) — all green**, one named test each:

| Gate | Invariant | Test |
|---|---|---|
| G1.1 | Identity: injected LoRA (B=0) forward == stock forward, ≤1e-6 | `test_injection_identity` |
| G1.2 | Merge parity: `forward(merge(m)) ≈ forward(m)`, ≤1e-5 | `test_merge_parity` |
| G1.3 | EMA alignment: student/teacher `named_parameters()` identical | `test_ema_zip_alignment` |
| G1.4 | Freeze audit: grads only on `lora_*`, heads, `mask_token` | `test_trainable_set` |
| G1.5 | Key stability: keys == stock ∪ `lora_*`; timm round-trip intact | `test_state_dict_keys` |
| G1.6 | Anchor-teacher sync with `lora_*` keys present | `test_gram_sync_lora` |
| G1.7 | Optimizer `blocks.N` layer-id grouping; no frozen params | `test_optim_groups` |

**Gate G2 (smoke training, 1×A100, ViT-B, 2×500 steps)** — LoRA injected into 24 modules per
backbone with **442,368 adapter parameters** (matching the design's ~0.44 M estimate), pretrained
remap loaded **162/162** tensors, run completed `Execution status: PASS`. Full G2.1–G2.6 audit
results are being appended as they land.

### Stack

`main` ← #84 (this module) ← #85 (wiring) ← #86 (tests). Review and merge in order.
Design docs are tracked in JIRA (epic TAO-2488) rather than in-repo.

### DIFF
--- a/nvidia_tao_pytorch/ssl/dinov3/model/lora.py
+++ b/nvidia_tao_pytorch/ssl/dinov3/model/lora.py
@@ -1,0 +1,367 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Key-preserving LoRA for the DINOv3 SSL backbone.
+
+Parameter-efficient continual pre-training: instead of updating all 86 M ViT-B backbone
+weights, freeze them and learn a rank-``r`` update ``dW = (alpha/r) * B @ A`` per targeted
+projection. See ``docs/plans/dinov3_lora_design.md``.
+
+**Key-preserving by design.** :class:`LoRALinear` subclasses ``nn.Linear`` and adopts the
+wrapped layer's existing ``weight``/``bias`` *Parameter objects*, so the backbone state-dict
+keys are exactly the stock keys plus ``lora_A``/``lora_B``. This is deliberately *not* the
+CLIP branch's ``.original``-nesting wrapper, which renames every wrapped weight to
+``...qkv.original.weight``. Four DINOv3 subsystems depend on stable backbone key names:
+
+* ``checkpoint_remap.timm_to_tao`` / ``convert_ssl_to_timm`` (timm <-> TAO translation),
+* ``CustomModelCheckpoint`` (strips ``student.backbone.``/``teacher.backbone.`` prefixes),
+* ``DinoV3PlModel._sync_gram_teacher`` (loads the teacher backbone into the plain anchor teacher),
+* ``restore_pretrained_weights`` (non-strict load against reference keys).
+
+**EMA safety.** ``DinoV2PlModel.update_teacher`` zips ``student.parameters()`` with
+``teacher.parameters()`` element-wise, so LoRA must be injected into the student *and* the
+EMA teacher with identical structure. Injection order here is deterministic
+(``nn.Module`` registration order), so the two zips stay aligned. The frozen anchor (Gram)
+teacher deliberately gets **no** LoRA -- it anchors to the pretrained weights, which under
+LoRA are precisely the frozen base weights.
+"""
+
+import math
+
+import torch
+from torch import nn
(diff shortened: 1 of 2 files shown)


nnagrajrao (commented):
Nice direction overall. I found one idempotence issue in the injection helper.

nvidia_tao_pytorch/ssl/dinov3/model/lora.py
```diff
@@ -0,0 +1,353 @@
...
+    injected = []
+    for block_index, block in _resolve_target_blocks(backbone, num_last_blocks):
+        for target in target_modules:
+            parent = getattr(block, _TARGET_PARENT[target], None)
+            if parent is None:
+                continue
+            child = getattr(parent, target, None)
+            if child is None:
+                continue
+            if isinstance(child, LoRALinear):
```
nnagrajrao, line 254: `inject_lora()` isn't actually idempotent right now. On the second call, the freeze loop above sets the existing `lora_A`/`lora_B` params to `requires_grad=False`, then this branch skips them, so they never get turned back on. The same thing happens if we inject qkv first and proj later. Can we freeze only non-LoRA params (or explicitly re-enable existing adapters) and add a double-injection test?

vpraveen-nv: Good catch — confirmed, and worse than it looks in review. I reproduced it before changing anything:

- second `inject_lora()` call: trainable tensors went **9 -> 1** (only `mask_token` survived)
- staged `qkv` then `proj`: **0** qkv adapters trainable, 4 proj adapters trainable

So a staged injection silently trained the heads plus half the adapters, with nothing in the logs to show it.

Fixed as you suggested — the freeze loop now skips LoRA parameters:

```python
for name, param in backbone.named_parameters():
    if not is_lora_key(name):
        param.requires_grad = False
```

and after injection every adapter present is explicitly re-enabled, so ordering can't matter. `test_injection_is_idempotent` in #86 covers both the repeat call and the staged `qkv`-then-`proj` case.

yuw-nv (approved):
LGTM

nnagrajrao (approved):

PR #85, opened 2026-08-06
[TAO-2489][Feature] DINOv3 LoRA: injection lifecycle and CLS preservation losses

vpraveen-nv:
Wires LoRA into the DINOv3 training flow and adds the CLS-token preservation losses.

**Injection point.** `train.py` calls `inject_lora_adapters()` between
`restore_pretrained_weights` and `trainer.fit` — after the load so the loader only ever sees
stock backbone keys, before `fit` so Lightning resume checkpoints and any distributed wrapping
see the final module tree.

**Both backbones.** LoRA goes into the student *and* the EMA teacher, because `update_teacher`
zips `student.parameters()` with `teacher.parameters()` element-wise and any asymmetry silently
corrupts the EMA. The student's backbone is then mirrored into the teacher: `lora_A` is randomly
initialized, so without that the teacher's adapter would not be an EMA of the student's at step 0.

**Anchor teacher.** It deliberately gets no LoRA — it anchors to the pretrained weights, which
under LoRA *are* the frozen base weights. It is now built when gram **or** preservation is
enabled, and `_sync_gram_teacher` filters `lora_*` keys and asserts the remaining base keys
cover it.

**One anchor forward.** `_extra_losses` runs the frozen teacher once and derives both terms from
it — the same output dict carries `x_norm_clstoken` alongside `x_norm_patchtokens`, so CLS
preservation costs no extra teacher pass.

**Guardrails**, asserted at the top of `_build_model` so they report their own reason rather than
tripping a downstream assert first: LoRA is rejected with `distill.enable`,
`gram.teacher_source='ema'`, and FSDP in v1.

JIRA: TAO-2489.

### Test evidence

`pytest tests/ssl_unit_test/dinov3`, measured against `main`:

| | result |
|---|---|
| baseline (`main`) | 75 passed, 4 skipped |
| with the full stack | **100 passed, 4 skipped** |

All 25 tests are additions; the same 4 skip for the same reason, so there are no regressions.
`pylint --rcfile .pylintrc` 10.00/10 on the changed sources; `pydocstyle` and `flake8` clean.

**Gate G1 (static invariants) — all green**, one named test each:

| Gate | Invariant | Test |
|---|---|---|
| G1.1 | Identity: injected LoRA (B=0) forward == stock forward, ≤1e-6 | `test_injection_identity` |
| G1.2 | Merge parity: `forward(merge(m)) ≈ forward(m)`, ≤1e-5 | `test_merge_parity` |
| G1.3 | EMA alignment: student/teacher `named_parameters()` identical | `test_ema_zip_alignment` |
| G1.4 | Freeze audit: grads only on `lora_*`, heads, `mask_token` | `test_trainable_set` |
| G1.5 | Key stability: keys == stock ∪ `lora_*`; timm round-trip intact | `test_state_dict_keys` |
| G1.6 | Anchor-teacher sync with `lora_*` keys present | `test_gram_sync_lora` |
| G1.7 | Optimizer `blocks.N` layer-id grouping; no frozen params | `test_optim_groups` |

**Gate G2 (smoke training, 1×A100, ViT-B, 2×500 steps)** — LoRA injected into 24 modules per
backbone with **442,368 adapter parameters** (matching the design's ~0.44 M estimate), pretrained
remap loaded **162/162** tensors, run completed `Execution status: PASS`. Full G2.1–G2.6 audit
results are being appended as they land.

### Stack

`main` ← #84 (this module) ← #85 (wiring) ← #86 (tests). Review and merge in order.
Design docs are tracked in JIRA (epic TAO-2488) rather than in-repo.

### DIFF
--- a/nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
+++ b/nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
@@ -37,12 +37,14 @@ from nvidia_tao_pytorch.ssl.dinov3.model.vit import DinoV3VisionTransformer, Swi
 from nvidia_tao_pytorch.ssl.dinov3.model.loss import GramLoss, ClsPreservationLoss
 from nvidia_tao_pytorch.ssl.dinov3.model.lora import (
     inject_lora,
+    is_lora_key,
     lora_parameter_report,
     strip_lora_keys,
 )
 from nvidia_tao_pytorch.ssl.dinov3.utils.checkpoint_remap import (
     TAO_ONLY_KEYS,
     fuse_timm_swiglu_fc1,
+    merge_lora_state_dict,
     timm_to_tao,
 )
 
@@ -265,7 +267,22 @@ class DinoV3PlModel(DinoV2PlModel):
         )
 
         injected = inject_lora(self.student.backbone, **kwargs)
-        inject_lora(self.teacher.backbone, **kwargs)
+        teacher_injected = inject_lora(self.teacher.backbone, **kwargs)
+
+        # An empty target set would otherwise freeze the whole backbone, mark LoRA as injected,
+        # and quietly train only the heads and mask_token -- a run that looks like LoRA and is
+        # not. Comparing the two lists also catches any future student/teacher topology
+        # divergence at injection time rather than as a silently misaligned EMA (gate G1.3).
+        assert injected, (
+            f"LoRA injected no modules (target_modules={target_modules}, "
+            f"num_last_blocks={lora_cfg.num_last_blocks}). model.lora.enable is set, so this "
+            "would train only the heads and mask_token while reporting a LoRA run."
+        )
+        assert injected == teacher_injected, (
+            "student and teacher received different LoRA injections "
+            f"(student={injected}, teacher={teacher_injected}); the EMA zip requires identical "
+            "parameter structure in both backbones."
+        )
 
         # Mirror the student's backbone (base + freshly initialized adapters) into the teacher
         # so both start identical; otherwise the teacher's random lora_A would make its EMA
@@ -652,6 +669,15 @@ class DinoV3PlModel(DinoV2PlModel):
             Tuple[dict, list]: ``(remapped, unmapped)`` where ``remapped`` is loadable into
             the backbone and ``unmapped`` lists source keys that had no shape-matching target.
         """
+        # Source carries adapters but the destination backbone cannot hold them (inference,
+        # export, or the start of a fresh run before injection): fold them into the base so the
+        # weights loaded are the *adapted* model. Without this the adapter keys are simply
+        # unmapped and the frozen base is silently used instead. When the destination is
+        # already LoRA-injected the adapters load normally and this does nothing.
+        if any(is_lora_key(k) for k in timm_state_dict) and \
+                not any(is_lora_key(k) for k in reference_state_dict):
+            timm_state_dict = merge_lora_state_dict(timm_state_dict)
+
         # Fuse split SwiGLU projections (fc1_g/fc1_x) into the fused fc1 for ViT-H+/7B before
         # the per-key shape match; plain-MLP (ViT-B/L) checkpoints are unaffected.
         timm_state_dict = fuse_timm_swiglu_fc1(timm_state_dict)
@@ -785,9 +811,31 @@ class DinoV3PlModel(DinoV2PlModel):
           preserving the global embedding geometry that k-NN/linear-probe/retrieval consume.
           Gated on ``preservation.enable``.
 
-        Both are logged unconditionally when active (even at weight 0) so they double as drift
-        diagnostics. Returns an empty list -- matching the DINOv2 base, hence unchanged
-        numerics -- when neither term is active.
+        **Input contract (important when reading these values).** The student tensors come from
+        the *masked* global crops the SSL step already computed, while the anchor runs on the
+        *unmasked* crops -- matching how DINOv3 applies Gram anchoring, and avoiding a second
+        student forward. Both terms are therefore a sum of two things: preservation of the
+        pretrained geometry (the intent) **and** sensitivity to iBOT masking. They are
+        consequently **not zero at step 0** even when the weights are identical, and they must
+        not be read as pure drift metrics.
+
+        Measured on ViT-B at init, where student and anchor weights are identical by
+        construction (LoRA is an identity at ``lora_B = 0``):
+
+        =====================  =========  =========  =========
+        condition              gram       cls_mse    cls_cos
+        =====================  =========  =========  =========
+        unmasked, eval          3.7e-09    4.3e-08    1.2e-07
+        30% masked, train       2.4e-02    5.8e-01    9.3e-01
+        =====================  =========  =========  =========
+
+        So the identity property is exact, but only observable with masking and stochastic
+        depth disabled -- which is how a sync/remap regression should be checked. During
+        training the value is dominated by masking and drop_path noise. For a true drift
+        measurement, run the anchor comparison periodically in eval mode on unmasked crops.
+
+        Both terms are logged unconditionally when active (even at weight 0). Returns an empty
+        list -- matching the DINOv2 base, hence unchanged numerics -- when neither is active.
 
         Args:
             **ctx: Context forwarded from ``DinoV2PlModel.student_forward`` (global/local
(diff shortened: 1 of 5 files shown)


nvidia_tao_pytorch/ssl/dinov3/scripts/train.py
```diff
@@ -69,6 +69,12 @@ def run_experiment(experiment_config, key):
         model.pretrained_weights = pretrained_path
         model.restore_pretrained_weights()
 
+    # LoRA injection sits between the pretrained load and `fit` on purpose: after the load so
+    # `restore_pretrained_weights` only ever sees stock backbone keys, and before `fit` so
+    # Lightning resume checkpoints and any distributed wrapping see the final module tree.
+    # No-op unless model.lora.enable is set.
+    model.inject_lora_adapters()
```
nnagrajrao, line 76: I think we're missing the other half of the lifecycle here. This is the only production call to inject or merge LoRA. `convert_ssl_to_timm()` passes `lora_A`/`lora_B` through and fails validation (or writes an unmerged state dict with validation off), while export and inference build a plain backbone and ignore the adapter keys. That means a trained LoRA checkpoint exports the frozen base model, not the adapted one. Can we wire those paths to merge `W + (alpha/r) * B @ A` and cover it with a nonzero-adapter checkpoint test?

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
```diff
@@ -91,16 +98,195 @@ def __init__(self, experiment_spec):
...
+            rank=lora_cfg.rank,
+            alpha=lora_cfg.alpha,
+            dropout=lora_cfg.dropout,
+            target_modules=target_modules,
+            num_last_blocks=lora_cfg.num_last_blocks,
+            freeze_base=True,
+        )
+
+        injected = inject_lora(self.student.backbone, **kwargs)
+        inject_lora(self.teacher.backbone, **kwargs)
```
nnagrajrao: Could we keep the return value from both injections and assert that the lists are non-empty and identical? With `target_modules=[]`, this currently freezes the whole backbone, marks LoRA as injected, and carries on training just the heads and mask token. It would also give us a runtime guard against a future student/teacher topology mismatch.

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
```diff
@@ -598,19 +836,36 @@ def _extra_losses(self, **ctx):
...
+            gram_loss = self.gram_loss(
+                student_backbone_global_output["x_norm_patchtokens"], teacher_patch_tokens
+            )
+            self._log_loss("losses/gram_loss", gram_loss)
+            losses.append(gram_cfg.w_gram * gram_loss)
+
+        if preservation_active:
+            # The CLS token is unaffected by the patch-grid pooling above, so it is read
+            # straight off the same anchor forward.
+            cls_mse, cls_cosine = self.cls_preservation_loss(
```
nnagrajrao, line 905: One design question: the student features here come from masked crops, but the anchor forward above is unmasked. That makes this partly a mask-invariance objective, and it won't be zero at step 0 even when the weights match. The PR already notes the nonzero measurement, so if this is intentional, can we state that contract clearly and stop describing G2.1 as zero-start? If the goal is only model-drift preservation, both sides need the same input/mask.

nnagrajrao (commented):
The training-side wiring looks sensible, but I think the checkpoint lifecycle is incomplete.

nvidia_tao_pytorch/ssl/dinov3/scripts/train.py
vpraveen-nv: You're right, and I'd deferred this too casually — thanks for pushing on it. Reproduced on a checkpoint with nonzero adapters:

- `convert(validate=True)` -> `ValueError: Converted state dict is not timm-compatible`
- `convert(validate=False)` -> **succeeded, writing 8 `lora_*` keys into a supposedly timm-format file**

The second is the dangerous one: a plausible-looking artifact that is actually the frozen base model.

Added `merge_lora_state_dict()`, folding `W + scaling * (B @ A)` and dropping the adapter keys, called from:

- `convert_ssl_to_timm`, before the timm translation
- `export`, before the plain backbone load, so the traced graph is stock DINOv3 topology
- the remap path generally, whenever the source carries adapters and the destination backbone cannot hold them — which also covers `inference`

A LoRA-injected destination still loads adapters normally, so resuming a LoRA run is unaffected, and full-fine-tune checkpoints are untouched.

One addition beyond your ask: the scale is now persisted as a `lora_scaling` buffer next to each adapter rather than re-derived from whichever spec is passed at merge time. Otherwise a mismatched `alpha` silently yields a wrongly-scaled backbone — the same failure mode you're flagging here.

`test_convert_merges_adapters_into_exported_weights` in #86 drives the real converter with nonzero A/B and asserts the exported weight equals `W + (alpha/r) * B @ A` with no `lora_*` surviving.

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
vpraveen-nv: Agreed on both counts — added:

```python
assert injected, (...)                      # empty target set would train only heads + mask_token
assert injected == teacher_injected, (...)  # student/teacher topology must match for the EMA zip
```

The `target_modules=[]` case was exactly as you describe: whole backbone frozen, `_lora_injected` set, training continues on the heads and `mask_token` while reporting a LoRA run. The list comparison also gives the runtime guard you wanted against a future student/teacher divergence, complementing the static `test_ema_zip_alignment` check.

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
vpraveen-nv: This is the sharpest comment of the set, and you've identified the real contract — it is partly a mask-invariance objective. I swept it on ViT-B to separate the causes:

| drop_path | mode | masking | gram | cls_mse | cls_cos |
|---|---|---|---|---|---|
| 0.0 | eval | none | 3.7e-09 | 4.3e-08 | **1.2e-07** |
| 0.0 | eval | 30% | 1.95e-02 | 2.93e-01 | 4.16e-01 |
| 0.4 | train | none | 1.86e-02 | 6.14e-01 | 9.25e-01 |
| 0.4 | train | 30% | 2.43e-02 | 5.75e-01 | 9.28e-01 |

So the weight-identity property is exact (1.2e-07 unmasked), but in a real training config the value is dominated by **stochastic depth** first and masking second.

I've taken your first option — state the contract, drop the "zero-start" framing — rather than masking the anchor, because the unmasked anchor is what DINOv3 does for Gram anchoring and matching it would need a second student forward. The `_extra_losses` docstring now spells this out and carries the table above, and `diagnostics/cls_anchor_cosine` is renamed **`cls_anchor_cosine_masked`** so it isn't read as pure drift. For an actual drift measurement the right instrument is a periodic eval-mode unmasked anchor comparison, which is now what the plan says.

`test_extra_losses_identity_through_the_real_path` in #86 enforces both halves: exact identity unmasked, and that masking materially moves the term — so if someone later makes the anchor see the same masks, that test fails and the docs get corrected with it.

vpraveen-nv:
Thanks for the review — all three findings reproduced, and all are fixed. Summary across the stack:

| Finding | PR | Status |
|---|---|---|
| `inject_lora` not idempotent | #84 | fixed; freeze skips LoRA params, adapters re-enabled after injection |
| convert/export/inference never merged adapters | #85 | fixed; `merge_lora_state_dict()` wired into all three |
| no guard on empty / mismatched injection | #85 | fixed; asserts non-empty and student == teacher |
| masked-vs-unmasked contract mis-described | #85 | contract documented with measured values; "zero-start" framing dropped; diagnostic renamed |
| key-stability test self-stripped adapters | #86 | replaced with an end-to-end converter test |
| preservation test only proved `f(x,x)==0` | #86 | renamed as a primitive; real-path test added |
| `requires_cuda` hid four gates from CPU CI | #86 | removed from nine tests |

Verification after the changes: **110 passed, 4 skipped** on GPU, **105 passed, 9 skipped** CPU-only (those four gates now actually run there), pylint 10.00/10, pydocstyle and flake8 clean.

Two of my own new tests failed on their first run and are fixed: the key-stability test didn't know about the new `lora_scaling` buffer, and the `_extra_losses` test omitted `use_custom_attention=false`, which the xformers path requires for an fp32 comparison.

The stack is rebased on `80f4b62f` and all four branches are re-pushed; `/build` re-run on each.

nnagrajrao (approved):

PR #86, opened 2026-08-06
[TAO-2489][Feature] DINOv3 LoRA: unit tests, HF checkpoint loading and SSL fixes (supersedes #87, #96, #97)

vpraveen-nv:
Consolidates four separate PRs into one, per discussion with the repo owners. Supersedes and
closes **#87**, **#96** and **#97**.

### Why one PR

The four changes were a stacked chain, and every merge auto-rebased the descendants, which
discarded their (~3 hour) `blossom-ci` results and forced a fresh run each time. Landing them
together avoids three more of those cycles.

### Revertability is preserved

The branch carries **exactly one commit per original PR**, so any single change can still be
reverted independently:

| Commit | Origin | Change |
|---|---|---|
| `92f444d9` | #86 | DINOv3 LoRA unit tests, gates G1.1–G1.7 |
| `f2b29cc2` | #87 | Load HuggingFace-format DINOv3 checkpoints |
| `95e3ea91` | #96 | Honour `attn_bias` in the SDPA fallback attention (2 commits squashed to 1) |
| `3ca5e5e6` | #97 | Honour `train.checkpoint_interval_unit='step'` |

Only #96 had multiple commits; those were squashed so each PR maps to a single revertable unit.

### Contents

- **Tests (#86)** — one test per Stage-1 invariant. Structural gates run on CPU, so G1.3/G1.4/G1.6/G1.7 are exercised in CPU-only CI too.
- **HF checkpoints (#87)** — the HF `DINOv3ViTModel` export holds the same weights as the timm release but stores attention as separate `q_proj`/`k_proj`/`v_proj`. Without conversion it matched **2 of 211** tensors and trained from an almost random backbone while still logging a successful load. Now 163/163, with `convert_hf_to_tao` refusing to convert if a q/k/v bias is nonzero rather than silently dropping weights.
- **SDPA `attn_bias` (#96)** — the non-xformers fallback ignored the attention bias.
- **`checkpoint_interval_unit` (#97)** — DINOv3 honoured only `'epoch'`; `'step'` was silently ignored.

### Verification on the combined branch

- `pytest tests/ssl_unit_test/dinov3 tests/ssl_unit_test/nvdinov2`: **136 passed, 4 skipped**
  (nvdinov2 included because #96 touches its attention layer)
- pylint 10.00/10 on all changed sources; pydocstyle and flake8 clean
- Each source PR's files were diffed against the combined branch to confirm nothing was lost;
  `pl_model.py` carries both #87's HF conversion and #97's interval handling

JIRA: TAO-2489, TAO-2492.

### DIFF
--- a/tests/ssl_unit_test/dinov3/test_lora.py
+++ b/tests/ssl_unit_test/dinov3/test_lora.py
@@ -1,0 +1,663 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""DINOv3 LoRA unit tests — the Stage-1 gate ladder (G1.1 – G1.7).
+
+Each test maps to one invariant from ``docs/plans/dinov3_lora_devbox_plan.md``:
+
+======  ==============================================================================
+G1.1    Identity: injected LoRA (B=0) forward == stock forward
+G1.2    Merge parity: forward(merge(m)) ~= forward(m) after random A/B
+G1.3    EMA alignment: student/teacher ``named_parameters()`` name lists identical
+G1.4    Freeze audit: ``requires_grad`` true only for ``lora_*``, heads, ``mask_token``
+G1.5    Key stability: backbone keys == stock keys + ``lora_*``; timm round-trip unaffected
+G1.6    Gram-teacher sync with a LoRA-injected teacher succeeds
+G1.7    Optimizer groups: LoRA params land in the right ``blocks.N`` layer-id groups
+======  ==============================================================================
+
+The backbone tests build a *tiny* ViT (2 blocks, 64-dim) on CPU in fp32 and disable
+``use_custom_attention``: the xformers ``memory_efficient_attention`` path hard-casts q/k/v to
+``.half()``, which would swamp the 1e-6 identity tolerance these gates are about.
+"""
+
+import pytest
+from omegaconf import OmegaConf
+import torch
+from torch import nn
+from pytorch_lightning.strategies.single_device import SingleDeviceStrategy
+
+from nvidia_tao_pytorch.config.dinov3.default_config import ExperimentConfig
+from nvidia_tao_pytorch.ssl.dinov3.model.lora import (
+    LoRALinear,
+    inject_lora,
+    merge_lora,
+    has_lora,
+    is_lora_key,
+    strip_lora_keys,
+    lora_parameter_report,
+)
+from nvidia_tao_pytorch.ssl.dinov3.model.pl_model import DinoV3PlModel
+from nvidia_tao_pytorch.ssl.dinov3.model.vit import DinoV3VisionTransformer
+from nvidia_tao_pytorch.ssl.dinov3.utils.checkpoint_remap import timm_to_tao, tao_to_timm
+
+# Tiny ViT so the whole gate ladder runs on CPU in seconds.
+TINY = dict(
+    img_size=32, patch_size=16, embed_dim=64, depth=2, num_heads=4,
+    init_values=1e-5, drop_path_schedule="uniform", num_classes=0,
+    drop_path_rate=0.0, register_tokens=2, use_custom_attention=False,
+)
+BATCH_SIZE = 2
+RANK = 4
+ALPHA = 8.0
+
+requires_cuda = pytest.mark.skipif(not torch.cuda.is_available(), reason="requires CUDA GPU")
+
+
+def _tiny_backbone(seed=0):
+    """Build a small fp32 DINOv3 ViT on CPU for the structural gates."""
+    torch.manual_seed(seed)
+    return DinoV3VisionTransformer(**TINY).eval()
+
+
+def _tiny_input():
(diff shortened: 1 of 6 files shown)


tests/ssl_unit_test/dinov3/test_lora.py
```diff
@@ -0,0 +1,549 @@
...
+@pytest.mark.ssl_unit
+def test_timm_remap_roundtrip_unaffected_by_lora():
+    """G1.5: the timm<->TAO key translation is untouched for base keys, and skips lora keys."""
+    backbone = _tiny_backbone()
+    inject_lora(backbone, rank=RANK, alpha=ALPHA)
+
+    for key in strip_lora_keys(backbone.state_dict()):
+        assert timm_to_tao(tao_to_timm(key)) == key, f"round-trip broke for {key}"
+
+    # LoRA keys are recognized as adapter keys, so convert can drop them before translation.
```
nnagrajrao: This test gives us a false sense of coverage: it strips the LoRA keys itself and never calls the converter or export path. Production code currently doesn't do the merge/strip step. Could we make this an end-to-end test with nonzero A/B and assert that the exported weight is `W + (alpha/r) * B @ A` and that no `lora_*` keys remain?

tests/ssl_unit_test/dinov3/test_lora.py
```diff
@@ -0,0 +1,549 @@
...
+        )
+
+
+# --------------------------------------------------------------------------------------
+# Preservation loss behaviour (the G2.1 zero-start property, checked statically)
+# --------------------------------------------------------------------------------------
+
+@pytest.mark.ssl_unit
+def test_cls_preservation_loss_zero_for_identical_inputs():
+    """Both CLS preservation terms are exactly 0 when student == anchor (gate G2.1 at step 0)."""
```
nnagrajrao, line 591: This only proves `loss_fn(x, x) == 0`. In the real `_extra_losses()` path, `x` isn't the same on both sides: the student saw masks and the anchor didn't. So this doesn't cover the step-0 gate. I'd either rename this as a primitive unit test or exercise the real path with the calibrated tolerance from the PR description.

tests/ssl_unit_test/dinov3/test_lora.py
```diff
@@ -0,0 +1,549 @@
...
+    config.model.lora.rank = RANK
+    config.model.lora.alpha = ALPHA
+    config.model.lora.dropout = 0.0
+    config.model.gram.enable = gram
+    config.model.gram.w_gram = 1.0 if gram else 0.0
+    config.model.preservation.enable = preservation
+    return config
+
+
+@requires_cuda
```
nnagrajrao: Do these structural tests need CUDA? They don't move the model or tensors to CUDA or run a forward, so this marker means G1.3/G1.4/G1.6/G1.7 disappear on CPU CI. I'd prefer making them CPU-safe (ideally with the tiny backbone) so these gates run everywhere.

nnagrajrao (commented):
The unit coverage is useful. A few of these checks should exercise the real integration path, though.

tests/ssl_unit_test/dinov3/test_lora.py
vpraveen-nv: Fair — it was asserting on a transformation the test itself performed, which proves nothing about production. Replaced with `test_convert_merges_adapters_into_exported_weights`, which calls the real `convert_ssl_to_timm` on a checkpoint with randomized A/B and asserts:

- every adapted weight equals `W + (alpha/r) * B @ A`, computed independently of the converter
- no `lora_*` key survives the export
- the result actually differs from the frozen base, so the test cannot pass vacuously

The production merge it exercises is the one added in #85.

tests/ssl_unit_test/dinov3/test_lora.py
vpraveen-nv: Correct — it only covered `loss_fn(x, x) == 0`. I did both things you offered:

- renamed it `test_cls_preservation_loss_primitive_zero_for_identical_inputs` so its scope is honest
- added `test_extra_losses_identity_through_the_real_path`, which drives the production `_extra_losses` with student and anchor identical by construction, asserts exact identity with masking and drop_path off, and asserts masking moves the term — the calibrated behaviour from the PR description, now enforced rather than described

tests/ssl_unit_test/dinov3/test_lora.py
vpraveen-nv: They don't — thanks, this was hiding four gates. I checked with `CUDA_VISIBLE_DEVICES=""`: `DinoV3PlModel` builds and injects fine on CPU, and the student/teacher name lists compare correctly, so the marker was superstition on my part.

Removed from nine tests, including all of G1.3/G1.4/G1.6/G1.7. A CPU-only run now reports **105 passed, 9 skipped**, where those gates previously didn't execute at all; the 9 remaining skips are pre-existing tests elsewhere in the directory that genuinely need a GPU. On GPU the suite is **110 passed, 4 skipped**.

nnagrajrao (approved):

nnagrajrao (approved):

james77777778 (approved):

yuw-nv (approved):

PR #87, opened 2026-08-06
[TAO-2489][Feature] DINOv3: load HuggingFace-format checkpoints

vpraveen-nv:
Lets `train.pretrained_model_path` accept a HuggingFace `DINOv3ViTModel` export, not just
the timm release.

### The bug this fixes is silent

The HF export holds the **same weights** as the timm one — verified bit-exact:
`cat([q_proj, k_proj, v_proj])` equals timm's fused `attn.qkv.weight` with **zero** difference,
as do `patch_embed` and `cls_token`. Only the serialization differs, in two ways the timm
remapper structurally could not express:

1. Different names throughout — `embeddings.*` / `layer.N.*` rather than bare / `blocks.N.*`.
2. Attention stored as **separate** `q_proj`/`k_proj`/`v_proj` rather than a fused `qkv`. That is
   a tensor transform, not a rename, so it needs its own pass (cf. `fuse_timm_swiglu_fc1`).

Under the timm rules alone exactly two keys coincided — `norm.weight` and `norm.bias` — so an HF
checkpoint loaded **2 of 211** tensors and training would proceed from an almost entirely random
backbone while `restore_pretrained_weights` still logged a successful load. Every downstream
retention number would have been meaningless, with nothing in the logs to say so.

### What changed

`is_hf_dinov3_state_dict` / `hf_to_tao` / `convert_hf_to_tao` in `checkpoint_remap.py`, detected
up front in `_remap_dinov3_state_dict`. Callers, specs and config are unchanged — pointing
`pretrained_model_path` at an HF directory now just works.

**On the bias question.** HF's config advertises `query_bias: true, value_bias: true`, while TAO
and timm build attention with `qkv_bias=False` — which looks like a genuine architecture
conflict. The shipped tensors are all zero (0 of 12 nonzero per projection), so dropping them is
lossless. `convert_hf_to_tao` **asserts** that rather than assuming it, and refuses to convert a
checkpoint whose q/k/v bias is nonzero instead of silently discarding weights.

### Test evidence

`pytest tests/ssl_unit_test/dinov3`: **107 passed, 4 skipped** (101 before this PR; 6 new tests
in `test_hf_checkpoint.py`). pylint 10.00/10, pydocstyle and flake8 clean.

End-to-end against the real checkpoint on the internal share, through the production load path:

| source | tensors | remapped | unmapped | coverage | residual missing |
|---|---|---|---|---|---|
| HF `dinov3-vitb16-pretrain-lvd1689m` | 211 | 163 | 0 | **163/163** | none |
| timm `vit_base_patch16_dinov3` | 162 | 162 | 0 | 162/163 | none |

- `max |HF − timm|` over shared tensors: **0.000e+00**
- forward feature parity: `x_norm_clstoken` cosine **0.99999994**, `x_norm_patchtokens` cosine
  **1.00000000**, max-abs **0.000e+00** on both

HF is in fact the *better* source: it ships `mask_token`, which timm omits and which TAO
otherwise initializes randomly for iBOT — the one key that shows up as missing on every timm load.

JIRA: TAO-2489.

**Stack:** `main` ← #84 ← #85 ← #86 ← this PR. Review and merge in order.

### DIFF
--- a/nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
+++ b/nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
@@ -43,7 +43,9 @@ from nvidia_tao_pytorch.ssl.dinov3.model.lora import (
 )
 from nvidia_tao_pytorch.ssl.dinov3.utils.checkpoint_remap import (
     TAO_ONLY_KEYS,
+    convert_hf_to_tao,
     fuse_timm_swiglu_fc1,
+    is_hf_dinov3_state_dict,
     merge_lora_state_dict,
     timm_to_tao,
 )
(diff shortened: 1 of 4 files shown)


nnagrajrao (approved):
lgtm!

nnagrajrao (approved):

vpraveen-nv:
Superseded by #86, which now carries this change as commit `f2b29cc2` — "Load HuggingFace-format DINOv3 checkpoints".

**Revertability is preserved** — #86 keeps exactly one commit per original PR, so this change can still be reverted on its own.

PR #88, opened 2026-08-06
[TAO-000][Bugfix] Unify CLIP logit calibration ownership

nnagrajrao:

Replace duplicate adapter-owned CLIP calibration with one canonical owner across every supported model family and every consumer:

- Hugging Face SigLIP2 trains its native `backbone.inner.logit_scale` and optional `logit_bias` directly.
- OpenCLIP reuses its native calibration without registering adapter duplicates.
- C-RADIO reuses native adaptor calibration when available and creates one local fallback only when the selected adaptor has none.
- `None` preserves learned source calibration; loss-family defaults are used only for a required missing value; explicit per-field overrides update the canonical parameter in place.
- Forward/loss, optimizer grouping, scale clamping, logging, validation and zero-shot post-processing, checkpointing, and export all resolve the same parameter objects.
- Bias-optional models stay biasless internally. The fixed export contract emits a zero bias tensor without adding a trainable parameter.
- The scale ceiling may be raised only by trusted native or successfully restored calibration. A fresh configuration override cannot ratchet its own ceiling.
- Obsolete duplicate-root, C-RADIO adaptor-alias, and one-element-vector checkpoint layouts are rejected with a targeted error. Per requested scope, no migration or compatibility shim is included.
- Inline config field descriptions document the revised `None` and override semantics. NVIDIA-TAO/tao-deploy#33 mirrors those descriptions; no standalone Markdown documentation is added.

TAO created a second calibration pair even when a pretrained model already owned learned calibration. For `google/siglip2-so400m-patch16-256`, training therefore ignored pretrained raw `logit_scale=4.6994` and `logit_bias=-15.9324` and used generic SigLIP values `2.3026` and `-10.0`. This materially changed the initial logits and fine-tuning behavior.

The common adapter contract should standardize access to calibration, not manufacture duplicate parameter ownership.

### Related issues

JIRA: TAO-000.

Companion deploy-schema documentation: NVIDIA-TAO/tao-deploy#33.

Yes. This is an intentional breaking cleanup of the affected model/checkpoint interface:

- Calibration now has one canonical state-dict path and uses scalar (zero-dimensional) parameters.
- Checkpoints and exported artifacts created with the previous duplicate-root, C-RADIO alias, or one-element-vector layouts are not load-compatible and must be regenerated. Automatic migration is intentionally not provided.
- A biasless adapter forward path may return three values instead of a manufactured fourth bias value. Export preserves its fixed output schema by returning a zero bias tensor.
- Raw OpenCLIP fallback construction now honors both vision- and text-tower freeze flags consistently.
- Config keys, field types, and serialized defaults are unchanged. The documented meaning of `None` now preserves source-owned pretrained calibration when available.

- TAO PyTorch Docker, directly affected calibration/adapter/script modules: **126 passed**.
- TAO PyTorch Docker, full non-GPU CLIP unit suite excluding the CUDA-dependent retrieval-metrics module: **378 passed**.
- A full CLIP-suite attempt reached only CUDA retrieval failures caused by the available host-driver/container NVML mismatch; no non-GPU test failed.
- Exact PR-range pre-commit checks passed: license header, pylint, pydocstyle, and flake8; pylint reported **10.00/10**.
- Ruff, Python bytecode compilation, `git diff --check`, DCO validation, and TruffleHog passed.
- Independent runtime/model, test-coverage, and CI/docs panel reviews: **3/3 READY**.
- Companion tao-deploy Docker config tests: **2 passed**, with its static checks passing.

Generated-by: OpenAI Codex

### Release note

```release-note
Changed CLIP calibration to use one model-owned logit scale and optional bias across training, validation, zero-shot processing, checkpointing, and export. Pretrained SigLIP2 and OpenCLIP calibration is preserved when overrides are unset.

Checkpoints using the previous duplicate calibration, C-RADIO adaptor-alias, or one-element-vector layout are not load-compatible and must be regenerated; automatic migration is not provided. Biasless adapter forwards may return three values, while export emits a zero bias tensor. Raw OpenCLIP fallback models now honor configured tower-freeze flags.
```

### Checklist

- [x] My commits are signed off per the DCO (`git commit -s`) — see `CONTRIBUTING.md`
- [x] The code follows the project's style, and lint and format checks pass locally
- [x] All applicable non-GPU tests pass locally
- [x] I updated configuration documentation (inline schema descriptions and NVIDIA-TAO/tao-deploy#33)
- [ ] I updated the version or changelog (release note and explicit breaking-change guidance are included here)
- [x] If this touches components that are optional to install, the imports remain guarded
- [x] No secrets, credentials, proprietary data, or customer data are included in this PR
- [x] I understand this contribution is licensed under the repository's license, and that my commit author name and email become permanently public once merged

### Notes for reviewers

- No Markdown or RST files are changed.
- No legacy checkpoint migration is included. The load guard only converts opaque strict-load failures into an actionable regeneration message.
- Review the mirrored tao-deploy field documentation in NVIDIA-TAO/tao-deploy#33 alongside this implementation.


nvidia_tao_pytorch/multimodal/clip/model/adapters/base.py
```diff
@@ -28,29 +36,49 @@ class BaseCLIPAdapter(nn.Module, ABC):
...
+    logit_scale_max = DEFAULT_MAX_LOGIT_SCALE
+
     def __init__(
         self,
-        logit_scale_init: float = 2.3026,
-        logit_bias_init: float = -10.0,
+        logit_scale_init=None,
+        logit_bias_init=None,
+        loss_type='siglip',
+        owns_logit_parameters=True,
```
vpraveen-nv, line 54: **Blocking / needs decision: this breaks every pre-PR CLIP checkpoint.** Removing the unconditional root `logit_scale`/`logit_bias` (plus the CRADIO `adaptor.*` alias removal and the `[1]` → 0-D normalization) means `train.py`'s `load_state_dict(strict=True)` and `load_from_checkpoint` both hard-fail on old checkpoints, with a cryptic missing/unexpected-key + shape-mismatch error.

I agree migration is reasonably out of scope — for SigLIP2 the trained values live in the removed root keys while `backbone.inner.*` holds stale pretrained values, so a naive `strict=False` remap would silently load wrong calibration. Failing loudly is the safer default. But please add (a) a catch-and-rethrow with a clear "checkpoint predates calibration unification" message, and (b) a release note.

nvidia_tao_pytorch/multimodal/clip/model/logit_calibration.py
```diff
@@ -0,0 +1,187 @@
...
+def _copy_override(parameter, value, name):
+    """Copy an override into a canonical parameter without replacing it."""
+    if value is None:
+        return
+    scalar = scalar_override(value, name)
+    with torch.no_grad():
+        parameter.fill_(scalar)
+
+
+def preserve_logit_scale_ceiling_(model, value=None):
```
vpraveen-nv, line 82: **Ceiling ratchet is one-way and user-triggerable.** This takes `max(log(100), existing_max, current_scale)` and runs at construction and every `on_train_start`. A typo'd huge `init_logit_scale` permanently raises the ceiling and turns the clamp into a no-op for that run — the clamp was previously the backstop against exactly this.

Consider ratcheting only from source-owned (pretrained/restored) values, not explicit user overrides — or at least warn when an override raises the ceiling above `log(100)`.

nvidia_tao_pytorch/multimodal/clip/model/clip.py
```diff
@@ -129,6 +131,27 @@ def build_model(experiment_config,
...
+            model,
+            logit_scale_init=init_logit_scale,
+            logit_bias_init=init_logit_bias,
+            loss_type=loss_type,
+            bias_required=False,
+        )
+        if freeze_vision:
+            model.visual.requires_grad_(False)
+            model.visual.eval()
+        if freeze_text:
```
vpraveen-nv, line 145: **New behavior, not a bugfix:** the raw open_clip fallback path previously ignored `freeze_vision_encoder`/`freeze_text_encoder` entirely. Honoring them is a good fix, but it changes training behavior for existing configs on this path (previously everything trained) and isn't mentioned in the PR description — please call it out there.

nvidia_tao_pytorch/multimodal/clip/model/pl_clip_model.py
```diff
@@ -451,7 +452,14 @@ def _backward(self, outputs, batch=None):
                 text_accessory_ids=metadata.get("text_accessory_ids"),
             )
         elif len(outputs) == 3:
```
vpraveen-nv, line 461: The 3-tuple is handled for plain `SigLipLoss`/`ClipLoss`, but the metadata masked-loss branch above still assumes a 4-tuple. Only reachable via an invalid config (masked SigLIP on a biasless CLIP source), but the failure would be a confusing unpack error — a cheap explicit guard ("masked SigLIP loss requires a model with logit_bias") would help.

nvidia_tao_pytorch/multimodal/clip/model/adapters/base.py
```diff
@@ -28,29 +36,49 @@ class BaseCLIPAdapter(nn.Module, ABC):
...
+            preserve_logit_scale_ceiling_(self)
+
+    def _create_local_logit_parameters(
+        self, logit_scale_init=None, logit_bias_init=None, loss_type='siglip'
+    ):
+        """Create one adapter-owned fallback calibration pair."""
+        fallback_scale, fallback_bias = default_logit_calibration(loss_type)
+        scale = fallback_scale if logit_scale_init is None else logit_scale_init
+        bias = fallback_bias if logit_bias_init is None else logit_bias_init
+        self.logit_scale = nn.Parameter(torch.tensor(scalar_override(
```
vpraveen-nv, line 74: Minor dtype asymmetry: `_new_scalar_parameter` in logit_calibration.py matches the source's dtype/device, but this local-fallback path uses bare `torch.tensor(...)` (fp32/CPU). `module.to()` fixes it later so it's not a bug today, but the inconsistency is a trap — consider routing both through the same helper.

nvidia_tao_pytorch/multimodal/clip/model/adapters/radio.py
```diff
@@ -175,19 +237,25 @@ def _log_parameters(self):
...
+        """Yield text parameters, excluding canonical calibration."""
         attr = self._TEXT_MODEL_ATTR.get(self.adaptor_name)
         if attr is None:
             return
-        prefix = f'adaptor.{attr}'
+        logit_param_ids = {
+            id(param) for _, param in self.named_logit_parameters()
+        }
+        prefix = (
+            f'radio_model.adaptors.{self.adaptor_name}.{attr}'
```
vpraveen-nv, line 252: Prefix change from `adaptor.<attr>.*` to `radio_model.adaptors.<name>.<attr>.*` is correct (matches real state-dict keys now that the alias is gone), but anything keying on the old names — tower LR schedules, logging filters, param-group name matching — should be audited.

nvidia_tao_pytorch/config/clip/default_config.py
```diff
@@ -74,19 +74,22 @@ class CLIPModelConfig:
     init_logit_scale: Optional[float] = FLOAT_FIELD(
         value=None,
         default_value=None,
-        description="Override for the initial logit scale (log-space). "
-                    "When None, automatically set from train.loss_type: "
-                    "2.3026 (SigLIP) or 2.6592 (CLIP). "
+        description="Optional raw logit-scale override. When None, preserve "
```
vpraveen-nv, line 77: Doc drift: tao-deploy's mirrored `nvidia_tao_deploy/config/multimodal/clip/default_config.py` still has the old description text for these two fields. Runtime is unaffected, but please file a follow-up to sync the deploy-side docs.

vpraveen-nv (commented):
minor comments.

nvidia_tao_pytorch/multimodal/clip/model/adapters/base.py
nnagrajrao: Addressed in 87b7fc3. A prefix-aware, rejection-only load hook now detects the obsolete duplicate-root keys, the former C-RADIO adaptor alias, and the one-element-vector scalar layout, then emits an explicit message that the checkpoint predates CLIP logit-calibration ownership unification and must be regenerated. Current canonical states still load strictly. Per the agreed scope, the hook does not migrate or mutate checkpoint state. The PR body and release note now mark this as an intentional breaking change and state that prior checkpoints/artifacts must be regenerated.

nvidia_tao_pytorch/multimodal/clip/model/logit_calibration.py
nnagrajrao: Addressed in 87b7fc3. Native source calibration is captured before any explicit override, and locally/config-created values cannot ratchet the ceiling. Lightning marks calibration trusted in on_load_checkpoint, while the manual pretrained-state path ratchets only after a successful strict load. Added regression coverage for native pretrained values, fresh explicit overrides, local Base/C-RADIO fallbacks, and restored values.

nvidia_tao_pytorch/multimodal/clip/model/clip.py
nnagrajrao: Called out explicitly in the user-facing changes and release note. The raw OpenCLIP fallback now honors both configured tower-freeze flags consistently, and the adapter/optimizer tests cover the resulting parameter partitioning.

nvidia_tao_pytorch/multimodal/clip/model/pl_clip_model.py
nnagrajrao: Investigated, and an explicit rejection guard is not needed: model outputs are normalized before this branch, so a biasless source reaches MetadataMaskedSigLipLoss with logit_bias=None rather than causing a tuple-unpack failure. That loss already accepts an optional bias. I added a direct regression test for the three-value masked-SigLIP path and verified it produces a finite loss.

nvidia_tao_pytorch/multimodal/clip/model/adapters/base.py
nnagrajrao: Agreed that the construction helpers are asymmetric. module.to() already puts this registered fallback parameter on the module dtype/device before use, so there is no observed runtime defect. I left this unchanged to keep the calibration-ownership bugfix minimal; consolidating all local scalar creation can be a separate cleanup if desired.

nvidia_tao_pytorch/multimodal/clip/model/adapters/radio.py
nnagrajrao: Audited. Optimizer grouping consumes the adapter vision/text/logit accessors and parameter identity; tower ownership is carried separately via _tower. Parameter names are used only for decay categorization, so the canonical state-dict prefix does not alter LR schedules or logging filters. The C-RADIO regression test asserts the radio_model.adaptors... canonical path, single optimizer inclusion, and strict current-state loading.

nvidia_tao_pytorch/config/clip/default_config.py
nnagrajrao: Addressed in companion NVIDIA-TAO/tao-deploy#33. It is a one-file schema-description update whose two description blocks match tao-pytorch byte-for-byte. TAO Deploy Docker config tests (2 passed) and static checks passed; no Markdown files are added.

nnagrajrao:
@vpraveen-nv Thank you for the detailed review. I addressed the actionable threads, added the missing regressions, amended and force-pushed the single commit as 87b7fc3, and replied to each inline comment with its disposition. The previous authorized /build predates this head. Could you please review the updates and re-run /build for 87b7fc3? The mirrored config descriptions are in companion NVIDIA-TAO/tao-deploy#33, which also needs an authorized /build if you have access there.

vpraveen-nv (approved):
LGTM.

yangchris11:
Ran a sanity check and compare the checkpoints, the TensorBoard validation tags, steps, and values are exactly equal.

 Both initialized calibration as logit_scale=109.88, logit_bias=-15.93.

PR #90, opened 2026-08-06
[TAO-000][Bugfix] Keep CLIP cosine LR checkpoint-safe

nnagrajrao:
TAO's CLIP cosine scheduler returned a `numpy.float64` and assigned it directly to optimizer parameter groups. Saving training state therefore serialized NumPy scalar metadata into `optimizer_states`, which is not accepted by the restricted `weights_only=True` checkpoint loader under the NumPy 2.x runtime.

### What changed

- Guarantees the documented built-in Python `float` return type for warmup, constant, linear, and cosine scheduler paths before values reach optimizer state.
- Adds an end-to-end parameterized regression using a NumPy scalar input and covering scheduler output, optimizer-state serialization, and `weights_only=True` loading.
- Keeps the test safe for parallel execution by avoiding mutation of process-global PyTorch serialization allowlists.

This is intentionally a producer-only fix. It does not expand safe globals or add compatibility handling for already-written affected checkpoints. Configuration and documentation interfaces are unchanged.

### Test evidence

- TAO Docker, `tests/multimodal_unit_test/clip/test_optimizer_utils.py`: **30 passed**.
- Focused warmup/constant/linear/cosine checkpoint regression under two xdist workers: **4 passed**.
- TAO Docker, CPU-capable CLIP unit suite from the original patch: **333 passed**, with 13 CUDA-only retrieval tests deselected.
- Exact PR-range pre-commit CI: license, pylint, pydocstyle, and flake8 passed; production pylint **10.00/10**.
- Python bytecode compilation, `git diff --check`, DCO validation, and TruffleHog passed.
- Independent runtime/model, test, and CI/docs panel reviews on the final head: **3/3 READY**.

JIRA: TAO-000.


yangchris11 (approved):
Looks good to me.

tests/multimodal_unit_test/clip/test_optimizer_utils.py
```diff
@@ -219,6 +219,36 @@ def test_cosine_scheduler(self):
...
+        parameter = nn.Parameter(torch.tensor(1.0))
+        optimizer = torch.optim.AdamW([parameter], lr=1e-4)
+        optimizer.param_groups[0]['lr'] = lr
+        checkpoint_path = tmp_path / 'cosine_lr.ckpt'
+        torch.save(
+            {'optimizer_states': [optimizer.state_dict()]},
+            checkpoint_path,
+        )
+
+        original_safe_globals = torch.serialization.get_safe_globals()
```
vpraveen-nv: Nit: this mutates process-global `torch.serialization` safe-globals state. The clear + restore-in-`finally` is correct for serial runs, but could race if this suite ever runs under `pytest-xdist` or with other tests that touch safe globals concurrently. Fine as-is if the suite is serial — just flagging in case parallelization is on the roadmap.

nvidia_tao_pytorch/multimodal/clip/utils/utils.py
```diff
@@ -260,7 +260,7 @@ def compute_lr(step, base_lr, warmup_steps, max_steps, scheduler='cosine'):
         return base_lr * max(1.0 - progress / total, 0.0)
 
     # cosine (default)
-    return 0.5 * (1 + np.cos(np.pi * progress / total)) * base_lr
+    return float(0.5 * (1 + np.cos(np.pi * progress / total)) * base_lr)
```
vpraveen-nv, line 263: Optional hardening: this covers the only branch that currently produces a `numpy.float64` (I checked — warmup/constant/linear are pure Python arithmetic). But a cast at the assignment site in `pl_clip_model.py` (`param_group['lr'] = float(compute_lr(...))`) — or a `float(...)` on all return paths here — would also guard any future scheduler branch from reintroducing the same defect. Non-blocking.

vpraveen-nv (commented):
Minor reviews.

tests/multimodal_unit_test/clip/test_optimizer_utils.py
nnagrajrao: Addressed in 92c4be8. The test no longer calls get_safe_globals, clear_safe_globals, or add_safe_globals; it relies on an explicit built-in-float assertion followed by an ordinary weights_only=True checkpoint load. Each parameterized case uses its own tmp_path, and the four cases pass under two xdist workers.

nvidia_tao_pytorch/multimodal/clip/utils/utils.py
nnagrajrao: Addressed in 92c4be8. compute_lr now enforces a built-in float on warmup, constant, linear, and cosine returns. The parameterized regression supplies np.float64 base_lr so every cast is a genuine before/after check, then verifies the optimizer checkpoint loads with weights_only=True.

nnagrajrao:
@vpraveen-nv I addressed both follow-up comments in 92c4be8 and replied inline with the validation details. The final head has 3/3 panel READY, 30/30 focused Docker tests, the four-case regression passing under xdist, and all local static checks passing. Since the previous authorized build predates this head, could you please re-run /build for 92c4be8?

vpraveen-nv:
LGTM. approved.

vpraveen-nv (approved):

PR #92, opened 2026-08-10
[TAO-0000][Bugfix] Log Visual ChangeNet FAR checkpoint metric

ramanathan831:
## Summary
Adds an in-training `val_far` metric (false-alarm rate at 100% recall over the validation set) computed each validation epoch for visual_changenet classify: the fraction of PASS samples scoring ≥ the minimum defect score. Logged so the checkpointer can monitor it (`enable_topk` + `monitor=val_far` + `mode=min` retains exactly the best-FAR epoch).

## Motivation
`val_loss` and the AOI deployment metric FAR@100%recall diverge: across 40 AutoML trials, per-config best-FAR epochs spread over the entire training run (epochs 9–99). Selecting checkpoints by `val_loss` silently discards the best-FAR epoch.

## Validation
Checkpoint-time val_far matches the external inference + analyze_kpi pipeline to within 2e-6 percentage points.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
+++ b/nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
@@ -232,9 +232,38 @@ class ChangeNetPlModel(TAOLightningModule):
         _ = self._forward_pass(batch)
         loss, siam_score = self._backward_G()
         self.val_metrics.update(siam_score, label)
+        # Accumulate raw scores/labels for the epoch-level FAR@100%recall metric.
+        if not hasattr(self, "_val_far_scores"):
+            self._val_far_scores, self._val_far_labels = [], []
+        self._val_far_scores.append(siam_score.detach().reshape(-1).float().cpu())
+        self._val_far_labels.append(label.detach().reshape(-1).cpu())
         self.log("val_loss", loss, on_step=False, on_epoch=True, prog_bar=True, sync_dist=True, batch_size=batch_size)
 
         return loss
+
+    def _compute_val_far(self):
+        """FAR at 100% recall over the accumulated validation scores.
+
+        Convention (matches AOIMetrics): score > threshold => predicted FAIL;
+        label 1 = FAIL (defect), 0 = PASS. With the strict `>` rule, the
+        lowest FAR achieving recall==1.0 is obtained at the largest candidate
+        threshold below min(defect scores):
+            val_far = #(pass_scores >= min(defect_scores)) / n_pass
+        Returns FAR percentage in [0, 100], or None if the accumulated batch
+        contains no defect or no pass samples (e.g. sanity check).
+        """
+        import torch as _torch
+        if not getattr(self, "_val_far_scores", None):
+            return None
+        scores = _torch.cat(self._val_far_scores)
+        labels = _torch.cat(self._val_far_labels)
+        defect = scores[labels == 1]
+        passed = scores[labels == 0]
+        if defect.numel() == 0 or passed.numel() == 0:
+            return None
+        min_defect = defect.min()
+        far = (passed >= min_defect).float().mean().item() * 100.0
+        return far
 
     def on_validation_epoch_end(self):
         """Validation epoch end.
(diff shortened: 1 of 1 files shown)


nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
```diff
@@ -232,10 +232,38 @@ def validation_step(self, batch, batch_idx):
         _ = self._forward_pass(batch)
         loss, siam_score = self._backward_G()
         self.val_metrics.update(siam_score, label)
+        # Accumulate raw scores/labels for the epoch-level FAR@100%recall metric.
+        if not hasattr(self, "_val_far_scores"):
+            self._val_far_scores, self._val_far_labels = [], []
+        self._val_far_scores.append(siam_score.detach().reshape(-1).float().cpu())
```
vpraveen-nv, line 242: **Multi-GPU correctness:** scores/labels are accumulated per rank, and `_compute_val_far` runs on each rank's shard only. `self.log("val_far", ..., sync_dist=True)` then *averages* the per-rank FARs — but FAR@100%recall depends on the global `min(defect_scores)`, which differs per rank, so the mean of per-rank FARs ≠ global FAR (the statistic is max/min-sensitive, averaging is biased). Consider `self.all_gather()` on the concatenated score/label tensors in `on_validation_epoch_end`, compute FAR once on the full set, and log with `sync_dist=False`.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
```diff
@@ -244,11 +272,19 @@ def on_validation_epoch_end(self):
 
         val_accuracy = self.val_metrics.compute()['total_accuracy'].item()
         val_fpr = self.val_metrics.compute()['false_alarm'].item()
+        val_far = self._compute_val_far()
+        self._val_far_scores, self._val_far_labels = [], []
+        if val_far is not None:
+            # Logged so ModelCheckpoint can monitor `val_far` (mode=min).
+            self.log("val_far", val_far, on_step=False, on_epoch=True,
```
vpraveen-nv: **Potential DDP deadlock:** `val_far` can be `None` on a rank whose shard happens to contain no defect (or no pass) samples while other ranks got both — then only some ranks enter this `self.log(..., sync_dist=True)` collective and the job hangs. The None/not-None decision must be globally consistent: compute after an `all_gather` (see comment above), or always log a sentinel value on every rank.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
```diff
@@ -244,11 +272,19 @@ def on_validation_epoch_end(self):
 
         val_accuracy = self.val_metrics.compute()['total_accuracy'].item()
         val_fpr = self.val_metrics.compute()['false_alarm'].item()
+        val_far = self._compute_val_far()
+        self._val_far_scores, self._val_far_labels = [], []
+        if val_far is not None:
+            # Logged so ModelCheckpoint can monitor `val_far` (mode=min).
```
vpraveen-nv: If `val_far` is `None` (e.g. a validation set with no defect samples), the metric is simply not logged that epoch — a `ModelCheckpoint(monitor="val_far")` will then raise `MisconfigurationException` on its first check. Consider always logging a well-defined worst-case (e.g. `100.0`) when the metric is undefined, or documenting that `monitor=val_far` requires both classes present in the val split.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
```diff
@@ -232,10 +232,38 @@ def validation_step(self, batch, batch_idx):
         _ = self._forward_pass(batch)
         loss, siam_score = self._backward_G()
         self.val_metrics.update(siam_score, label)
+        # Accumulate raw scores/labels for the epoch-level FAR@100%recall metric.
+        if not hasattr(self, "_val_far_scores"):
```
vpraveen-nv: Nit: lazy-init via `hasattr` in the hot loop works but is fragile (state survives across fit/validate calls until epoch end resets it). `on_validation_epoch_start` is the idiomatic place to reset these buffers.

vpraveen-nv (commented):
minor comments.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
ramanathan831: Addressed in eccf520a: validation splits without both classes now log the explicit 100.0 worst-case sentinel, so ModelCheckpoint always sees val_far.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
ramanathan831: Addressed in eccf520a: all ranks participate in the gather before the class-presence decision, so the result is globally consistent and cannot diverge into a DDP deadlock.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
ramanathan831: Addressed in eccf520a: validation scores and labels are gathered across ranks, FAR is computed once from the global validation set, and val_far is logged without a second averaging collective.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
ramanathan831: Addressed in eccf520a: FAR buffers are initialized and reset in on_validation_epoch_start instead of lazily in the validation hot loop.

vpraveen-nv (approved):
All comments addressed — verified the global all_gather FAR computation, the sentinel logging, and the epoch-start buffer reset in the updated diff. LGTM.

PR #95, opened 2026-08-11
[TAO-2511][Feature] Support frozen DINOv3 backbones in Visual ChangeNet

nnagrajrao:

Add frozen DINOv3 backbone support to the existing Visual ChangeNet classification and segmentation paths with minimal changes to the AOI architecture.

- Supports ViT-S, S+, B, L, H+, and 7B through the existing ViT adapter and task heads.
- Keeps the DINOv3 ViT frozen while the AOI adapter and classification or segmentation head remain trainable.
- Gives an explicit converted TAO checkpoint precedence; when no checkpoint is configured, frozen DINOv3 loads matching timm/Hugging Face weights.
- Routes explicit euclidean checkpoints through the existing AOI parser and state-dict loader, including actionable rejection of raw SSL checkpoints.
- Adds the 7B channel map and 40-block adapter partition.
- Preserves existing C-RADIO behavior and defaults.

Example:

```yaml
model:
  backbone:
    type: vit_large_dinov3
    pretrained_backbone_path: /data/pretrained_models/dinov3_vit_l_backbone.safetensors
    freeze_backbone: true
```

Set `pretrained_backbone_path: null` to use the matching timm/Hugging Face pretrained weights.

DEFT AOI needs to reuse DINOv3 SSL features without updating the backbone. This keeps the established Visual ChangeNet task-level fine-tuning design while allowing either a converted TAO DINOv3 checkpoint or the compatible public timm/Hugging Face weights.

### Related issues

JIRA: TAO-2511.

Companion changes are in tao-core and tao-skill-bank on the same TAO-2511 branch.

Yes. Visual ChangeNet users can select all six DINOv3 variants and freeze the ViT. Existing backbone types, C-RADIO behavior, and defaults are unchanged.

- Required TAO image, focused DINOv3 and config parity suite: **12 passed**.
- Required TAO image, broader Visual ChangeNet model suite: **66 passed, 2 skipped**.
- Exact PR-range CI hooks: license, pylint, pydocstyle, and flake8 passed.
- Confirmed timm 1.0.26 exposes all six configured model names.
- Independent architecture, correctness, and MR-readiness panel: **3/3 READY**.
- `git diff --check` and no-Markdown checks passed.

Generated-by: OpenAI Codex

### Release note

```release-note
Visual ChangeNet classification and segmentation now support frozen DINOv3 ViT-S, S+, B, L, H+, and 7B backbones using converted TAO checkpoints or compatible timm/Hugging Face pretrained weights.
```

### Checklist

- [x] My commits are signed off per the DCO (`git commit -s`)
- [x] The code follows the project style, and lint and format checks pass locally
- [x] All applicable tests pass locally
- [x] I updated inline configuration documentation and docstrings
- [ ] A version or changelog update is not required for this patch
- [x] Optional imports remain guarded
- [x] No secrets, credentials, proprietary data, or customer data are included
- [x] I understand the contribution is licensed under the repository license

### Notes for reviewers

No Markdown files are changed. Start with `cv/backbone_v2/dino_v3.py`, then the two Visual ChangeNet model constructors and the focused regression test. The change deliberately reuses existing AOI adapters and heads instead of introducing a parallel task package.

### DIFF
--- a/nvidia_tao_pytorch/cv/backbone_v2/dino_v3.py
+++ b/nvidia_tao_pytorch/cv/backbone_v2/dino_v3.py
@@ -16,6 +16,67 @@ import timm
 
 from nvidia_tao_pytorch.cv.backbone_v2 import BACKBONE_REGISTRY
 from nvidia_tao_pytorch.cv.backbone_v2.backbone_base import BackboneBase
+from nvidia_tao_pytorch.core.utils.ptm_utils import load_pretrained_weights
+
+
+_CONVERSION_HINT = (
+    "Run `dinov3 convert -e <dinov3_spec.yaml> "
+    "convert.checkpoint=<ssl_checkpoint> "
+    "convert.output_path=<converted_backbone.safetensors>` first."
+)
+
+
+def validate_dinov3_checkpoint(state_dict, allow_partial=False):
+    """Validate that a DINOv3 state dict uses the downstream timm layout.
+
+    TAO DINOv3 SSL checkpoints use a different key layout and must pass through
+    the existing `dinov3 convert` task before a downstream task consumes them.
+    Hugging Face `timm/*dinov3*` checkpoints and converted TAO checkpoints
+    both use the accepted timm layout.
+
+    Args:
+        state_dict (Mapping): DINOv3 state dict to validate.
+        allow_partial (bool): Permit incomplete timm-format state dicts while
+            retaining raw SSL checkpoint rejection.
+
+    Raises:
+        ValueError: If the state dict is empty, is a raw TAO SSL checkpoint, or
+            does not contain the required timm DINOv3 backbone keys.
+    """
+    if not state_dict and not allow_partial:
+        raise ValueError(f"DINOv3 checkpoint is empty. {_CONVERSION_HINT}")
+
+    keys = []
+    for key in state_dict:
+        normalized = key
+        for prefix in ("vit.", "inner."):
+            if normalized.startswith(prefix):
+                normalized = normalized[len(prefix):]
+                break
+        keys.append(normalized)
+
+    raw_ssl_markers = ("mask_token", "register_tokens")
+    is_raw_ssl = (
+        any(key in raw_ssl_markers for key in keys) or
+        any(key.endswith((".ls1.gamma", ".ls2.gamma")) for key in keys) or
+        any(key.startswith(("teacher.", "student.")) for key in keys)
+    )
+    if is_raw_ssl:
+        raise ValueError(
+            "Raw TAO DINOv3 SSL checkpoints are not accepted by downstream "
+            f"task backbones. {_CONVERSION_HINT}"
+        )
+
+    required_keys = ("cls_token", "patch_embed.proj.weight")
+    if allow_partial:
+        return
+    if not all(key in keys for key in required_keys) or not any(
+        key.startswith("blocks.0.") for key in keys
+    ):
+        raise ValueError(
+            "DINOv3 checkpoint is not a compatible timm-format backbone. "
+            f"{_CONVERSION_HINT}"
+        )
 
 
 class DINOV3Wrapper(BackboneBase):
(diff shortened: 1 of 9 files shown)


nvidia_tao_pytorch/cv/visual_changenet/classification/models/changenet.py
```diff
@@ -465,11 +465,13 @@ def __init__(
                     resolution=output_shape[0],
                     activation_checkpoint=activation_checkpoint,
                     use_summary_token=use_summary_token,
+                    pretrained=freeze_backbone and not pretrained_backbone_path,
```
vpraveen-nv: With `freeze_backbone: true` and no checkpoint, this triggers an HF Hub download at model-build time — on every DDP rank concurrently, and the DINOv3 HF weights are license-gated (HF token required). In an air-gapped training container this fails with a raw `HfHubHTTPError`. Worth a try/except with a clear message (pointing at `pretrained_backbone_path` + `dinov3 convert`), and possibly a rank-0 guard + barrier for the download?

nvidia_tao_pytorch/cv/visual_changenet/classification/models/changenet.py
```diff
@@ -425,7 +425,7 @@ def __init__(
 
         freeze_at = None
         if freeze_backbone:
-            if pretrained_backbone_path is None:
+            if not pretrained_backbone_path and 'dinov3' not in self.model_name:
```
vpraveen-nv, line 429: Minor UX asymmetry: for DINOv3 with no path, `freeze_backbone: true` downloads HF weights while `freeze_backbone: false` trains from random init — same knob, different weight source, and both fields are `automl_enabled`. A rank-0 log line stating which weight source was used when the HF fallback engages would help debugging sweeps.

nvidia_tao_pytorch/cv/backbone_v2/dino_v3.py
```diff
@@ -117,6 +173,7 @@ def load_state_dict(self, state_dict, **kwargs):
             state_dict (dict): a dict containing parameters and persistent buffers.
             **kwargs: Additional arguments passed to `nn.Module.load_state_dict`.
         """
+        validate_dinov3_checkpoint(state_dict)
```
vpraveen-nv: Two scope questions: (1) this tightens `load_state_dict` for every `DINOV3Wrapper` consumer, not just ChangeNet — e.g. a deliberately-partial state dict under `strict=False` now gets rejected. Intended? (2) The registry's `pretrained_backbone_path` flow (`_consumes_pretrained_path`) goes through timm's `checkpoint_path` loader and bypasses this validation, so a raw SSL checkpoint there still fails with timm's cryptic key mismatch. Worth covering `_load_dino_v3` too?

vpraveen-nv (approved):
LGTM — the 7B wiring (40-block partition, 4096 channel maps) checks out against timm, the raw-SSL rejection matches the converter's remap table, and the C-RADIO/FAN paths are untouched. A few non-blocking questions inline. One sequencing note: please land the tao-core companion (NVIDIA-TAO/tao-core#31) in lockstep so schema validation doesn't reject `vit_7b_dinov3` in the window between merges.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/changenet.py
nnagrajrao: Addressed in 8cbbcf6. The shared DINOv3 creator now catches timm/Hugging Face failures and raises actionable guidance for HF_TOKEN, a local pretrained_backbone_path, and dinov3 convert. I did not add a global-rank barrier because caches may be node-local in multi-node jobs; Hugging Face cache locking coordinates ranks that share a cache, while each node must still be able to populate its own cache. Covered by the focused fallback-error test; the 13 focused and 94 broader tests pass.

nvidia_tao_pytorch/cv/backbone_v2/dino_v3.py
nnagrajrao: Addressed both points in 8cbbcf6. strict=False now permits intentional partial timm-layout loads while still rejecting raw SSL markers. The explicit registry path now loads, validates, and then strictly applies the local checkpoint instead of bypassing validation through checkpoint_path. Regression tests cover both behaviors.

nvidia_tao_pytorch/cv/visual_changenet/classification/models/changenet.py
nnagrajrao: Addressed in 8cbbcf6. Classification and segmentation now emit a rank-zero log when a frozen DINOv3 backbone selects timm/Hugging Face weights, and both paths have regression assertions.

PR #96, opened 2026-08-11
[TAO-2492] Honour attn_bias in the SDPA fallback attention (silently corrupts SSL training on Hopper/Blackwell; also ~8x faster)

vpraveen-nv:
## Summary

`MemoryEfficientAttention.forward` passes `attn_bias` to xformers, but called
`self._fallback_attention(q, k, v)` **without it**.

The SSL blocks concatenate their crop list into a single sequence via
`get_attn_bias_and_cat`, so one "batch" row holds every image's tokens end to end. The
block-diagonal `attn_bias` is the **only** thing preventing image *i*'s tokens from attending
to image *j*'s. Dropping it blends the entire batch together.

## Impact

**Every DINOv3 and NVDINOv2 training run with batch > 1 on any GPU where xformers custom
attention is unavailable** — Hopper (where it is force-disabled) and Blackwell (by design).

A100 is unaffected because it takes the xformers path, which carries the mask. That is why this
survived workstation testing: the gates all pass there.

The failure is silent. Losses look normal, checkpoints look normal, and every *relative* check
(identity, frozen-base bit-identity, key preservation, drift-vs-anchor) still passes, because
both sides of each comparison come from the same corrupted model.

## Evidence — correctness

Measured on H100 against timm's `vit_base_patch16_dinov3.lvd1689m` loaded from the **same
checkpoint** (weights bit-identical: 162/162 tensors, `allclose` atol 1e-6):

| batch | CLS cosine vs timm — before | after |
|---|---|---|
| 1 | 1.000000 | 1.000000 |
| 2 | 0.657706 | **1.000000** |
| 4 | 0.368179 | **1.000000** |
| 8 | 0.192695 | **1.000000** |

| | before | after | reference |
|---|---|---|---|
| ImageNet k-NN top-1 (100 classes, k=20, chance 1.00%) | **4.00%** | **82.40%** | timm 82.40% |
| Feature pairwise cosine at batch 8 | 0.8750 | 0.0568 | timm 0.0568 |

The clearest symptom: at batch 8 the pairwise feature cosine was **0.875 for both random noise
and real photographs** — identical. The features had stopped depending on their input.

## Evidence — throughput (the fix is also much faster)

Attention over the concatenated sequence is O(N²) in the total token count rather than per
crop, so the missing mask was not only destroying features, it was paying a large multiple of
the necessary compute to do it:

| 8×H100, multi-crop SSL config | rate |
|---|---|
| before | 0.13 it/s |
| after | **0.99–1.18 it/s** |

On a full-scale training run the end-to-end speedup measured ~18×.

## Implementation note — split, don't materialize

The obvious fix, materializing `attn_bias` into a dense additive `[1, heads, N, N]` mask, is
correct but **unusable in training**: it allocates precisely what xformers' structured
`BlockDiagonalMask` exists to avoid. At DINOv3 multi-crop sequence lengths that mask is tens of
GiB — I tried it first and it OOMed four training jobs within minutes.

This PR instead splits the concatenated sequence back into per-crop blocks and attends within
each. Block-diagonal attention *is* independent attention per block, and `split` is the
documented inverse of `get_attn_bias_and_cat`, so it is mathematically identical and allocates
nothing extra.

## Verification

* bit-exact against timm at batch 1/2/4/8
* full multi-crop training path: no OOM, sustained 0.99–1.18 it/s

## Note for reviewers

Two changes met to create this, neither wrong alone: the SDPA fallback added for the fp16 NaN
guard, and the auto-disable of xformers on newer architectures. The fallback simply never
carried the mask its xformers counterpart did.

A regression test would be worth adding, but it must compare against a reference implementation
on **real images**. A random-input test does *not* catch this — noise produces near-uniform
attention that hides the missing mask, and a block-by-block activation diff on random input
matched to 1e-6 while the model was thoroughly broken.

### DIFF
--- a/nvidia_tao_pytorch/ssl/dinov3/model/layers/attention.py
+++ b/nvidia_tao_pytorch/ssl/dinov3/model/layers/attention.py
@@ -58,7 +58,12 @@ class RoPEMemoryEfficientAttention(MemoryEfficientAttention):
             # the shared SDPA-based fallback (base MemoryEfficientAttention._fallback_attention) so
             # both SSL families get the same numerically-safe, fused compute; RoPE has already been
             # applied to q/k above. See bug 6460915.
-            x = self._fallback_attention(q, k, v)
+            #
+            # attn_bias is forwarded: the block concatenates its crop list into one sequence,
+            # so without the block-diagonal mask every image attends to every other image in
+            # the batch. That silently destroyed features on Hopper, where this path is the
+            # only one available (bug 6459926 forces xformers off).
+            x = self._fallback_attention(q, k, v, attn_bias)
 
         x = x.reshape(B, N, C)
         x = self.proj(x)
--- a/nvidia_tao_pytorch/ssl/dinov3/model/layers/attention.py
+++ b/nvidia_tao_pytorch/ssl/dinov3/model/layers/attention.py
@@ -61,8 +61,8 @@ class RoPEMemoryEfficientAttention(MemoryEfficientAttention):
             #
             # attn_bias is forwarded: the block concatenates its crop list into one sequence,
             # so without the block-diagonal mask every image attends to every other image in
-            # the batch. That silently destroyed features on Hopper, where this path is the
-            # only one available (bug 6459926 forces xformers off).
+            # the batch, which silently corrupts features wherever this fallback is the only
+            # available path.
             x = self._fallback_attention(q, k, v, attn_bias)
 
         x = x.reshape(B, N, C)
--- a/nvidia_tao_pytorch/ssl/nvdinov2/model/layers/attention.py
+++ b/nvidia_tao_pytorch/ssl/nvdinov2/model/layers/attention.py
@@ -27,20 +27,17 @@ class MemoryEfficientAttention(Attention):
         GPUs, where ``use_custom_attention`` is force-disabled for both SSL families. See bug 6460915
         and the MR !652 SDPA review.
 
-        ``attn_bias`` MUST be honoured. The SSL blocks concatenate their crop list into a
-        single sequence via ``get_attn_bias_and_cat`` -- so one "batch" row holds every image's
-        tokens end to end -- and the block-diagonal mask is the only thing preventing image i's
+        ``attn_bias`` must be honoured. The SSL blocks concatenate their crop list into a
+        single sequence via ``get_attn_bias_and_cat``, so one "batch" row holds every image's
+        tokens end to end, and the block-diagonal mask is the only thing preventing image i's
         tokens from attending to image j's. Dropping it silently blends the whole batch
-        together: features stop depending on their own image, and the damage grows with batch
-        size. Measured on H100 against timm's reference DINOv3, CLS cosine was 1.000 at batch 1,
-        0.658 at 2, 0.368 at 4 and 0.193 at 8, and ImageNet k-NN fell from 82.4% to 4.0%.
+        together, and the damage grows with batch size.
 
-        The mask is applied by **splitting**, not by materializing. A dense
+        The mask is applied by splitting rather than materializing. A dense
         ``[1, heads, N, N]`` mask is what xformers' structured ``BlockDiagonalMask`` exists to
-        avoid: for DINOv3 training (2 global crops at 512 px + 4 local at 112 px, batch 16) the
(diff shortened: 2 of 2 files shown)


nvidia_tao_pytorch/ssl/dinov3/model/layers/attention.py
```diff
@@ -58,7 +58,12 @@ def forward(self, x, attn_bias=None, use_custom_attention=True, rope=None):
             # the shared SDPA-based fallback (base MemoryEfficientAttention._fallback_attention) so
             # both SSL families get the same numerically-safe, fused compute; RoPE has already been
             # applied to q/k above. See bug 6460915.
-            x = self._fallback_attention(q, k, v)
+            #
+            # attn_bias is forwarded: the block concatenates its crop list into one sequence,
+            # so without the block-diagonal mask every image attends to every other image in
+            # the batch. That silently destroyed features on Hopper, where this path is the
+            # only one available (bug 6459926 forces xformers off).
```
james77777778: should we remove the internal bug number?

nvidia_tao_pytorch/ssl/nvdinov2/model/layers/attention.py
```diff
@@ -27,20 +27,51 @@ def _fallback_attention(self, q, k, v):
...
+        tokens end to end -- and the block-diagonal mask is the only thing preventing image i's
+        tokens from attending to image j's. Dropping it silently blends the whole batch
+        together: features stop depending on their own image, and the damage grows with batch
+        size. Measured on H100 against timm's reference DINOv3, CLS cosine was 1.000 at batch 1,
+        0.658 at 2, 0.368 at 4 and 0.193 at 8, and ImageNet k-NN fell from 82.4% to 4.0%.
+
+        The mask is applied by **splitting**, not by materializing. A dense
+        ``[1, heads, N, N]`` mask is what xformers' structured ``BlockDiagonalMask`` exists to
+        avoid: for DINOv3 training (2 global crops at 512 px + 4 local at 112 px, batch 16) the
+        concatenated sequence is ~36k tokens, so the dense mask would be ~55 GiB and OOMs
```
james77777778: i think we could remove these numbers to keep docstrings cleaner.

nvidia_tao_pytorch/ssl/nvdinov2/model/layers/attention.py
```diff
@@ -27,20 +27,51 @@ def _fallback_attention(self, q, k, v):
         GPUs, where ``use_custom_attention`` is force-disabled for both SSL families. See bug 6460915
         and the MR !652 SDPA review.
 
+        ``attn_bias`` MUST be honoured. The SSL blocks concatenate their crop list into a
+        single sequence via ``get_attn_bias_and_cat`` -- so one "batch" row holds every image's
+        tokens end to end -- and the block-diagonal mask is the only thing preventing image i's
+        tokens from attending to image j's. Dropping it silently blends the whole batch
+        together: features stop depending on their own image, and the damage grows with batch
+        size. Measured on H100 against timm's reference DINOv3, CLS cosine was 1.000 at batch 1,
+        0.658 at 2, 0.368 at 4 and 0.193 at 8, and ImageNet k-NN fell from 82.4% to 4.0%.
```
james77777778: same.

james77777778 (commented):
minor comments. overall, LGTM. thanks for the perf fix!

vpraveen-nv:
Thanks for the review!

Both addressed in `325b890c`:

- **Measurement figures** removed from the `_fallback_attention` docstring (per-batch CLS cosine, k-NN percentages, token counts, GiB). Agreed these read as campaign-specific noise in the source — the reasoning stays in the docstring, the numbers live in the PR description above.
- **Internal bug id** removed from the DINOv3 comment; it now just says the fallback is the only available path there.

One thing I left alone deliberately: both files already referenced `bug 6460915` / `MR !652` before this PR, and I kept those untouched to hold the diff to your review's scope. If the preference is that no internal ids appear in the source at all, say the word and I'll strip them in a follow-up (or here, if you'd rather it be one change).

Re-running CI on the new head below.

vpraveen-nv:
Superseded by #86, which now carries this change as commit `95e3ea91` — "Honour attn_bias in the SDPA fallback attention".

**Revertability is preserved** — #86 keeps exactly one commit per original PR (the two commits here were squashed into one), so this change can still be reverted on its own.

PR #106, opened 2026-08-19
[TAO-1796][Feature] Enable LoRA fine-tuning for CLIP with per-tower modes

yangchris11:

This change adds configurable per-tower adaptation for CLIP models.

  Key changes:

  - Add `frozen`, `full`, and `lora` adaptation modes independently for the vision and text towers.
  - Make `mode` the sole per-tower interface, with `frozen`, `full`, and `lora` as the supported values.
  - Migrate the VideoCLIP schema and example specifications from tower-level `enabled` to explicit modes.
  - Allow logit calibration parameters to be trained independently through `train_logit_calibration`.
  - Inject LoRA adapters only into the requested final transformer blocks and projection modules.
  - Validate `num_last_blocks` against the selected tower depth.
  - Preserve tower-specific optimizer groups and learning rates.
  - Support checkpoint save/restore and merging LoRA weights into the base model for export.
  - Fail before mutating model state when LoRA is requested for an unsupported `nn.MultiheadAttention`-based OpenCLIP tower.
  - Document that OpenCLIP currently supports `full` and `frozen` modes; OpenCLIP LoRA support will be handled separately.

  Example for SigLIP2 vision-tower LoRA:

  ```yaml
  peft:
    enabled: true
    train_logit_calibration: true
    vision:
      mode: lora
      target_modules: [q_proj, k_proj, v_proj, out_proj]
      num_last_blocks: 3
      rank: 8
      alpha: 16
      dropout: 0.05
    text:
      mode: frozen
  ```

  Pre-release VideoCLIP LoRA configurations using tower-level `enabled` must replace `enabled: true` with `mode: lora` and `enabled: false` with `mode: frozen`. This interface has not appeared in a released TAO version.

  - CLIP fine-tuning previously exposed only per-tower LoRA enablement. It could not express hybrid strategies such as fully fine-tuning one tower while
    applying LoRA to, or freezing, the other tower.

  - The new modes provide one unambiguous per-tower interface without retaining an unreleased legacy compatibility layer.
  - Additional validation prevents silent no-op training caused by invalid block depths or unmatched projection names.
  - OpenCLIP uses fused nn.MultiheadAttention projections that are incompatible with the current nn.Linear LoRA wrapper. The change now detects this
    architecture before modifying model state and directs users to the supported full and frozen modes.

  - A full end-to-end experiment was conducted covering:
      - LoRA training
      - Checkpoint creation and restoration
      - Evaluation
      - Adapter merging
      - Model export

### Related issues

TAO-1796 and https://github.com/NVIDIA-TAO/tao-core/pull/36

Yes.

  Previously, each tower supported only an `enabled` Boolean for LoRA. Users can now select one of the following modes independently:

  - `frozen`: freeze the tower.
  - `full`: fully fine-tune the tower.
  - `lora`: inject and train LoRA adapters.
  Logit calibration training can also be controlled independently using `train_logit_calibration`.

  The pre-release tower-level `enabled` interface has been removed. Configurations using it must migrate as follows:

  - `enabled: true` → `mode: lora`
  - `enabled: false` → `mode: frozen`

  The top-level `peft.enabled` switch remains unchanged. No released TAO configuration requires migration.

  Invalid LoRA block depths now produce a clear configuration error instead of implicitly adapting all blocks. OpenCLIP LoRA requests now fail early with
  an actionable message; OpenCLIP users should select `full` or `frozen` until fused-attention LoRA support is added.

Focused LoRA tests:

  ```console
  $ CUDA_VISIBLE_DEVICES='' pytest -q \
      tests/multimodal_unit_test/clip/test_lora.py \
      tests/multimodal_unit_test/clip/test_lora_peft.py

  76 passed
  ```

  Broader GPU-independent CLIP unit suite:

  ```console
  $ CUDA_VISIBLE_DEVICES='' pytest -q tests/multimodal_unit_test/clip \
      --ignore=tests/multimodal_unit_test/clip/test_retrieval_metrics.py

  460 passed
  ```

  Full VideoCLIP unit suite, including shared LoRA compatibility:

  ```console
  $ CUDA_VISIBLE_DEVICES='' pytest -q \
      tests/multimodal_unit_test/video_clip

  93 passed, 1 skipped
  ```

  Static validation:

  ```console
  $ pre-commit run --files $(git diff --name-only origin/main...HEAD)

  TruffleHog secret scan...................................................Passed
  dependency change guard..............................(no files to check)Skipped
  license header...........................................................Passed
  pylint...................................................................Passed
  pydocstyle...............................................................Passed
  flake8...................................................................Passed
  ```

  Additional manual verification:

  - Injected eight LoRA adapters into a real, reduced-size Hugging Face `Siglip2Model`.
  - Verified SigLIP2 forward and backward execution and gradients for all LoRA parameters.
  - Verified LoRA merging preserved SigLIP2 output exactly.
  - Verified a real OpenCLIP `ResidualAttentionBlock` is rejected before its structure or parameter trainability is changed.
  - Verified CLIP and VideoCLIP towers default to explicit `frozen` mode.
  - Verified the migrated VideoCLIP LoRA specification composes successfully with its structured schema.
  - Verified missing, `legacy`, `"???"`, and other unsupported modes are rejected.
  - Verified invalid block depths and unmatched target modules fail with actionable errors.

  GPU retrieval tests were not executed successfully on this host because its NVIDIA driver is older than the installed PyTorch CUDA runtime; those tests
  are unrelated to this change.

Generated-by: OpenAI Codex (GPT-5)

### Release note

```release-note
Added configurable per-tower `frozen`, `full`, and `lora` modes for CLIP; tower-level `enabled` configurations must migrate to `mode`, unsupported OpenCLIP LoRA requests now fail early, and `num_last_blocks` values exceeding the selected tower depth must be reduced or set to `0`.
```

### Checklist

- [X] My commits are signed off per the DCO (`git commit -s`) — see `CONTRIBUTING.md`
- [X] I have read the contributing guidelines
- [X] The code follows the project's style, and lint and format checks pass locally
- [X] I added or updated tests covering this change
- [X] All tests pass locally
- [X] I added or updated documentation (README, docstrings, docs pages, examples)
- [X] I updated the version, changelog, or migration notes if this change requires it
- [X] If this touches components that are optional to install, the imports are guarded
      so the package still works without them (reviewers: please verify this)
- [X] No secrets, credentials, proprietary data, or customer data are included in this PR
- [X] I understand this contribution is licensed under the repository's license, and that
      my commit author name and email become permanently public once merged

### Notes for reviewers

The main implementation is in `nvidia_tao_pytorch/multimodal/clip/model/lora.py`.

  Please focus on:

  - Per-tower `frozen`, `full`, and `lora` mode resolution.
  - Mode-only tower configuration and migration of the shared VideoCLIP consumer.
  - Schema parity with NVIDIA-TAO/tao-core#36.
  - Validation occurring before model parameters or modules are mutated.
  - Optimizer partitioning and logit-calibration handling.
  - LoRA checkpoint and merge behavior.

  OpenCLIP LoRA is intentionally out of scope because its fused `nn.MultiheadAttention` projections require a dedicated adapter. OpenCLIP `full` and
  `frozen` tower modes remain supported.

### DIFF
--- a/nvidia_tao_pytorch/config/clip/default_config.py
+++ b/nvidia_tao_pytorch/config/clip/default_config.py
@@ -113,30 +113,19 @@ class CLIPModelConfig:
 class CLIPLoRATargetConfig:
     """Adaptation configuration for a single encoder tower (vision or text).
 
-    Controls how a tower is adapted. ``mode: legacy`` preserves existing
-    enabled-only YAMLs by resolving ``enabled: true`` to ``lora`` and
-    ``enabled: false`` to ``frozen``.
+    Controls whether a tower is frozen, fully trainable, or adapted with LoRA.
     """
 
     mode: str = STR_FIELD(
-        value="legacy",
-        default_value="legacy",
-        valid_options="legacy,frozen,full,lora",
+        value="frozen",
+        default_value="frozen",
+        valid_options="frozen,full,lora",
         description=(
-            "Tower adaptation mode. 'legacy' resolves from the deprecated "
-            "enabled field; 'frozen' leaves the tower fixed; 'full' trains "
-            "all tower parameters; 'lora' trains only injected LoRA parameters."
+            "Tower adaptation mode. 'frozen' leaves the tower fixed; 'full' "
+            "trains all tower parameters; 'lora' trains only injected LoRA "
+            "parameters."
         ),
         display_name="Adaptation Mode",
-    )
-    enabled: Optional[bool] = BOOL_FIELD(
-        value=None,
-        default_value=None,
-        description=(
-            "Deprecated legacy switch. With mode='legacy', true selects LoRA "
-            "and false selects frozen. Omit it when using an explicit mode."
-        ),
-        display_name="Enabled (Legacy)",
     )
     target_modules: List[str] = LIST_FIELD(
         arrList=["q_proj", "k_proj", "v_proj", "out_proj"],
--- a/nvidia_tao_pytorch/config/clip/default_config.py
+++ b/nvidia_tao_pytorch/config/clip/default_config.py
@@ -111,18 +111,23 @@ class CLIPModelConfig:
 # =============================================================================
 @dataclass
 class CLIPLoRATargetConfig:
-    """LoRA configuration for a single encoder tower (vision or text).
+    """Adaptation configuration for a single encoder tower (vision or text).
 
     Controls which transformer blocks and attention modules receive
     low-rank adapters. When enabled, only the LoRA parameters in the
     selected blocks are trainable; all other backbone parameters are frozen.
     """
 
-    enabled: bool = BOOL_FIELD(
-        value=False,
-        default_value=False,
-        description="Enable LoRA adaptation for this encoder tower.",
-        display_name="Enabled",
+    mode: str = STR_FIELD(
+        value=MISSING,
+        default_value=MISSING,
+        valid_options="frozen,full,lora",
+        description=(
+            "Tower adaptation mode. 'frozen' leaves the tower fixed; 'full' "
+            "trains all tower parameters; 'lora' trains only injected LoRA "
+            "parameters. Required when PEFT is enabled."
+        ),
+        display_name="Adaptation Mode",
     )
     target_modules: List[str] = LIST_FIELD(
         arrList=["q_proj", "k_proj", "v_proj", "out_proj"],
@@ -187,6 +192,15 @@ class CLIPPEFTConfig:
         valid_options="lora",
         description="PEFT method. Currently only 'lora' is supported.",
         display_name="Method",
+    )
+    train_logit_calibration: bool = BOOL_FIELD(
+        value=True,
+        default_value=True,
+        description=(
+            "Train logit_scale and optional logit_bias while PEFT is enabled. "
+            "Defaults to True to preserve existing LoRA behavior."
+        ),
+        display_name="Train Logit Calibration",
     )
     vision: CLIPLoRATargetConfig = DATACLASS_FIELD(
         CLIPLoRATargetConfig(),
--- a/nvidia_tao_pytorch/config/clip/default_config.py
+++ b/nvidia_tao_pytorch/config/clip/default_config.py
@@ -141,10 +141,11 @@ class CLIPLoRATargetConfig:
     target_modules: List[str] = LIST_FIELD(
         arrList=["q_proj", "k_proj", "v_proj", "out_proj"],
         default_value=["q_proj", "k_proj", "v_proj", "out_proj"],
-        description="Module name substrings to target for LoRA injection. "
+        description="Module leaf names to target for LoRA injection. "
                     "SigLIP2: 'q_proj', 'k_proj', 'v_proj', 'out_proj'. "
-                    "RADIO/OpenCLIP: 'qkv', 'proj' (fused attention). "
-                    "Matching is by substring on the module name within each block.",
+                    "RADIO: 'qkv', 'proj' (fused attention). OpenCLIP uses "
+                    "nn.MultiheadAttention and does not support LoRA mode yet; "
+                    "use mode 'full' or 'frozen'.",
         display_name="Target Modules",
     )
     num_last_blocks: int = INT_FIELD(
--- a/nvidia_tao_pytorch/config/clip/default_config.py
+++ b/nvidia_tao_pytorch/config/clip/default_config.py
@@ -113,21 +113,30 @@ class CLIPModelConfig:
 class CLIPLoRATargetConfig:
     """Adaptation configuration for a single encoder tower (vision or text).
 
-    Controls which transformer blocks and attention modules receive
-    low-rank adapters. When enabled, only the LoRA parameters in the
-    selected blocks are trainable; all other backbone parameters are frozen.
+    Controls how a tower is adapted. ``mode: legacy`` preserves existing
+    enabled-only YAMLs by resolving ``enabled: true`` to ``lora`` and
+    ``enabled: false`` to ``frozen``.
     """
 
     mode: str = STR_FIELD(
-        value=MISSING,
-        default_value=MISSING,
-        valid_options="frozen,full,lora",
+        value="legacy",
+        default_value="legacy",
+        valid_options="legacy,frozen,full,lora",
         description=(
-            "Tower adaptation mode. 'frozen' leaves the tower fixed; 'full' "
-            "trains all tower parameters; 'lora' trains only injected LoRA "
-            "parameters. Required when PEFT is enabled."
+            "Tower adaptation mode. 'legacy' resolves from the deprecated "
+            "enabled field; 'frozen' leaves the tower fixed; 'full' trains "
+            "all tower parameters; 'lora' trains only injected LoRA parameters."
         ),
         display_name="Adaptation Mode",
+    )
+    enabled: Optional[bool] = BOOL_FIELD(
+        value=None,
+        default_value=None,
+        description=(
+            "Deprecated legacy switch. With mode='legacy', true selects LoRA "
+            "and false selects frozen. Omit it when using an explicit mode."
+        ),
+        display_name="Enabled (Legacy)",
     )
     target_modules: List[str] = LIST_FIELD(
         arrList=["q_proj", "k_proj", "v_proj", "out_proj"],
--- a/nvidia_tao_pytorch/multimodal/clip/model/lora.py
+++ b/nvidia_tao_pytorch/multimodal/clip/model/lora.py
@@ -122,8 +122,11 @@ def inject_lora(model, peft_config):
                 )
             target_blocks = blocks if num_last_blocks == 0 else blocks[-num_last_blocks:]
             start_index = len(blocks) - len(target_blocks)
+            available_linear_leaves = set()
             for offset, block in enumerate(target_blocks):
                 for name, module in list(block.named_modules()):
+                    if isinstance(module, nn.Linear):
+                        available_linear_leaves.add(name.rsplit('.', 1)[-1])
                     if not _match_target(name, tower_config.target_modules):
                         continue
                     parent_path, _, attribute = name.rpartition('.')
@@ -141,8 +144,11 @@ def inject_lora(model, peft_config):
         )
         if mode == 'lora' and tower_trainable_params[tower_name] == 0:
             raise ValueError(
-                f"PEFT {tower_name} mode='lora' injected zero modules. "
-                "Check target_modules and num_last_blocks for this backbone."
+                f"PEFT {tower_name} mode='lora' injected zero modules (no LoRA adapters): "
+                f"target_modules {list(tower_config.target_modules)} matched no "
+                f"nn.Linear modules. Available leaves: "
+                f"{sorted(available_linear_leaves)}. Check target_modules and "
+                "num_last_blocks for this backbone."
             )
 
     if _config_value(peft_config, 'train_logit_calibration', True):
--- a/nvidia_tao_pytorch/multimodal/clip/model/lora.py
+++ b/nvidia_tao_pytorch/multimodal/clip/model/lora.py
@@ -98,6 +98,30 @@ def _enable_lora_parameters(module):
     module.lora_B.requires_grad = True
 
 
+def _preflight_lora_towers(model, towers, resolved_modes):
+    """Reject unsupported LoRA architectures before changing model state."""
+    for tower_name, tower_config in towers:
+        if resolved_modes[tower_name] != 'lora':
+            continue
+        blocks = list(model.get_encoder_blocks(tower_name))
+        num_last_blocks = tower_config.num_last_blocks
+        if not 0 <= num_last_blocks <= len(blocks):
+            raise ValueError(
+                f"PEFT {tower_name}.num_last_blocks must be between 0 "
+                f"(all blocks) and {len(blocks)}, got {num_last_blocks}."
+            )
+        target_blocks = blocks if num_last_blocks == 0 else blocks[-num_last_blocks:]
+        if any(
+            isinstance(module, nn.MultiheadAttention)
+            for block in target_blocks for module in block.modules()
+        ):
+            raise NotImplementedError(
+                f"PEFT {tower_name} mode='lora' is not supported for "
+                "nn.MultiheadAttention-based towers (OpenCLIP). Use "
+                "mode='full' or mode='frozen'."
+            )
+
+
 def inject_lora(model, peft_config):
     """Apply explicit per-tower PEFT modes and inject LoRA where requested."""
     if not _config_value(peft_config, 'enabled', False):
@@ -107,6 +131,7 @@ def inject_lora(model, peft_config):
         ('text', _config_value(peft_config, 'text')),
     )
     resolved_modes = {name: resolve_tower_mode(config, name) for name, config in towers}
+    _preflight_lora_towers(model, towers, resolved_modes)
     for parameter in model.parameters():
         parameter.requires_grad = False
 
@@ -122,11 +147,6 @@ def inject_lora(model, peft_config):
         else:
             blocks = list(model.get_encoder_blocks(tower_name))
             num_last_blocks = tower_config.num_last_blocks
-            if not 0 <= num_last_blocks <= len(blocks):
-                raise ValueError(
-                    f"PEFT {tower_name}.num_last_blocks must be between 0 "
-                    f"(all blocks) and {len(blocks)}, got {num_last_blocks}."
-                )
             target_blocks = blocks if num_last_blocks == 0 else blocks[-num_last_blocks:]
             start_index = len(blocks) - len(target_blocks)
             available_linear_leaves = set()
--- a/nvidia_tao_pytorch/multimodal/clip/model/lora.py
+++ b/nvidia_tao_pytorch/multimodal/clip/model/lora.py
@@ -114,12 +114,13 @@ def inject_lora(model, peft_config):
                 parameter.requires_grad = True
         else:
             blocks = list(model.get_encoder_blocks(tower_name))
-            if tower_config.num_last_blocks > len(blocks):
+            num_last_blocks = tower_config.num_last_blocks
+            if not 0 <= num_last_blocks <= len(blocks):
                 raise ValueError(
-                    f"PEFT {tower_name}.num_last_blocks={tower_config.num_last_blocks} "
-                    f"exceeds the {len(blocks)} available transformer blocks."
+                    f"PEFT {tower_name}.num_last_blocks must be between 0 "
+                    f"(all blocks) and {len(blocks)}, got {num_last_blocks}."
                 )
-            target_blocks = blocks if tower_config.num_last_blocks == 0 else blocks[-tower_config.num_last_blocks:]
+            target_blocks = blocks if num_last_blocks == 0 else blocks[-num_last_blocks:]
             start_index = len(blocks) - len(target_blocks)
             for offset, block in enumerate(target_blocks):
                 for name, module in list(block.named_modules()):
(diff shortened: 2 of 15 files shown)


nvidia_tao_pytorch/multimodal/clip/model/lora.py
```diff
@@ -31,235 +13,206 @@
...
+    if not _config_value(peft_config, 'enabled', False):
+        return None
+    towers = (
+        ('vision', _config_value(peft_config, 'vision')),
+        ('text', _config_value(peft_config, 'text')),
     )
+    resolved_modes = {name: resolve_tower_mode(config, name) for name, config in towers}
+    _preflight_lora_towers(model, towers, resolved_modes)
+    for parameter in model.parameters():
+        parameter.requires_grad = False
```
nnagrajrao, line 275: Is train.grad_checkpointing=true expected to work with LoRA here? SigLIP2 uses reentrant HF checkpointing; after these base params are frozen, the checkpoint inputs no longer require grad and the LoRA weights may receive no gradients. Could we use non-reentrant checkpointing or enable input grads?

nvidia_tao_pytorch/multimodal/clip/model/lora.py
```diff
@@ -31,235 +13,206 @@
...
-
-        Returns:
-            The original nn.Linear with merged weights.
-        """
+        """Fold the LoRA residual into and return the wrapped linear layer."""
         with torch.no_grad():
-            # delta_W = B @ A * scaling, shape (out_features, in_features)
-            delta = (self.lora_B @ self.lora_A) * self.scaling
-            self.original.weight.add_(delta)
+            self.original.weight.add_((self.lora_B @ self.lora_A) * self.scaling)
```
nnagrajrao: Could we persist or validate the LoRA scale in the checkpoint? Export rebuilds the wrapper from the current spec, so changing alpha with the same rank still loads cleanly and this merge silently uses the new scale.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
```diff
@@ -31,235 +13,206 @@
...
+            raise ValueError(
+                f"PEFT {tower_name} mode='lora' injected zero modules (no LoRA adapters): "
+                f"target_modules {list(tower_config.target_modules)} matched no "
+                f"nn.Linear modules. Available leaves: "
+                f"{sorted(available_linear_leaves)}. Check target_modules and "
+                "num_last_blocks for this backbone."
+            )
 
+    if _config_value(peft_config, 'train_logit_calibration', True):
+        for _, parameter in model.named_logit_parameters():
```
vpraveen-nv: **Blocking:** `inject_lora` is shared — `multimodal/video_clip/model/pl_video_clip_model.py` imports and calls it for InternVideo2 PEFT, and the video_clip adapter base does **not** implement `named_logit_parameters()` (it only has plain `logit_scale`/`logit_bias` attributes). The previous code used a getter/attribute fallback precisely so both model families worked; this unconditional call (here and in the `logit_trainable_params` stat below) will raise `AttributeError` for every video_clip LoRA run. Either restore a fallback for models without the method, or add `named_logit_parameters()` to the video_clip adapter base in this PR — and please run `tests/multimodal_unit_test/video_clip` too (the PR test log covers only the clip suite).

nvidia_tao_pytorch/multimodal/clip/model/lora.py
```diff
@@ -31,235 +13,206 @@
...
-                f"adapters were injected: target_modules "
-                f"{list(tower_cfg.target_modules)} matched no nn.Linear in the "
-                f"last {len(target_blocks)} of {num_blocks} blocks. The "
-                f"backbone is already frozen at this point, so training would "
-                f"proceed with nothing trainable. nn.Linear leaf names "
-                f"available in those blocks: {sorted(seen_linear_leaves)}. "
-                f"(Fused attention, e.g. nn.MultiheadAttention's in_proj, is a "
-                f"Parameter rather than a Linear submodule and cannot be "
-                f"targeted by name.)"
+                f"PEFT {tower_name}.num_last_blocks must be between 0 "
```
vpraveen-nv, line 179: This is a behavior change worth being deliberate about: previously `num_last_blocks >= len(blocks)` silently meant "adapt all blocks"; now anything above the tower depth raises. I agree failing loudly is better, but (a) can you confirm no shipped experiment specs (including `video_clip/experiment_specs/experiment_spec_lora.yaml`) rely on an oversized value, and (b) since the depth differs per backbone, consider whether `num_last_blocks > depth` → clamp-with-warning would be friendlier for users switching backbones with one spec. If we keep the hard error, the release note should call out the migration explicitly.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
```diff
@@ -31,235 +13,206 @@
...
-
-        if tower_cfg.num_last_blocks == 0 or tower_cfg.num_last_blocks >= num_blocks:
-            target_blocks = blocks
-        else:
-            target_blocks = blocks[-tower_cfg.num_last_blocks:]
+def resolve_tower_mode(tower_config, tower_name):
+    """Resolve a tower mode, including enabled-only legacy configurations."""
+    mode = _config_value(tower_config, 'mode', LEGACY_TOWER_MODE)
+    enabled = _config_value(tower_config, 'enabled')
+    if mode in (None, '???', LEGACY_TOWER_MODE):
```
vpraveen-nv: Matching the literal `'???'` string couples this to an OmegaConf serialization detail. Prefer `omegaconf.OmegaConf.is_missing(tower_config, 'mode')` (or compare against `omegaconf.MISSING`) so this doesn't silently break if the missing-value representation changes, and so plain-dataclass/dict configs aren't surprised by a magic string.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
```diff
@@ -31,235 +13,206 @@
...
-            continue
 
-        blocks = model.get_encoder_blocks(tower_name)
-        num_blocks = len(blocks)
-
-        if tower_cfg.num_last_blocks == 0 or tower_cfg.num_last_blocks >= num_blocks:
-            target_blocks = blocks
-        else:
-            target_blocks = blocks[-tower_cfg.num_last_blocks:]
+def resolve_tower_mode(tower_config, tower_name):
```
vpraveen-nv, line 125: This feature has never shipped in a release (CLIP LoRA landed on main in #94 last week; no release branch contains it), so there are no external `enabled:` configs to preserve. Rather than carrying a permanent legacy layer — the `mode: legacy` sentinel, the `Optional` `enabled` field, and the mode-vs-enabled conflict validation — consider removing `enabled` outright and making `mode` the only interface, migrating `VideoCLIPLoRATargetConfig` to `mode` in the same pass (it's the only live consumer of the legacy fallback, and it also came in #94). If we do keep the compat layer for in-flight internal branches, it should be short-lived with an explicit removal target, not a silent permanent path.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
```diff
@@ -31,235 +13,206 @@
...
+                        available_linear_leaves.add(name.rsplit('.', 1)[-1])
+                    if not _match_target(name, tower_config.target_modules):
+                        continue
+                    parent_path, _, attribute = name.rpartition('.')
+                    parent = block.get_submodule(parent_path) if parent_path else block
+                    if isinstance(module, nn.Linear):
+                        module = LoRALinear(module, tower_config.rank, tower_config.alpha, tower_config.dropout)
+                        setattr(parent, attribute, module)
+                        injected.append((tower_name, f'block[{start_index + offset}].{name}'))
+                    if isinstance(module, LoRALinear):
```
vpraveen-nv: Question on the idempotent re-injection: which code path calls `inject_lora` twice on the same model (resume? re-instantiation?)? Two observations: (1) on the second call the stats semantics silently change — `injected_modules` is empty while `lora_params` still counts the existing adapters — so any consumer comparing those two fields gets inconsistent views between first and second call; (2) a repeated injection with a *different* rank/alpha config is silently ignored for already-wrapped modules. If double-injection is a real path, log a warning when existing `LoRALinear`s are re-enabled; if it isn't, consider raising instead.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
```diff
@@ -1,25 +1,7 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
 
-"""LoRA (Low-Rank Adaptation) utilities for CLIP models.
-
-Provides lightweight LoRA injection and merging for CLIP adapters.
-Works with both separate (q_proj, k_proj, ...) and fused (qkv) attention
-projections across SigLIP2, OpenCLIP, and RADIO backbones.
-"""
+"""LoRA injection and merge utilities for CLIP encoder projections."""
```
vpraveen-nv: Roughly half the deletions in this file (and in `adapters/base.py`) are docstring/comment removal — the Args/Returns/Raises sections on `LoRALinear`, `inject_lora`, and `merge_lora`, plus the explanatory comments, are gone. That's a docs regression unrelated to the feature and it makes the functional diff much harder to review. Please restore the API docstrings (they can mention the new modes), and keep cosmetic reformatting out of feature PRs.

nvidia_tao_pytorch/config/clip/default_config.py
```diff
@@ -111,26 +111,41 @@ class CLIPModelConfig:
...
+        default_value="legacy",
+        valid_options="legacy,frozen,full,lora",
+        description=(
+            "Tower adaptation mode. 'legacy' resolves from the deprecated "
+            "enabled field; 'frozen' leaves the tower fixed; 'full' trains "
+            "all tower parameters; 'lora' trains only injected LoRA parameters."
+        ),
+        display_name="Adaptation Mode",
+    )
+    enabled: Optional[bool] = BOOL_FIELD(
```
vpraveen-nv: If we drop the legacy `enabled` field per the comment in `lora.py` (the feature never shipped, so there's nothing to stay compatible with), this `Optional[bool] = BOOL_FIELD(value=None)` question disappears along with the field. If it stays: (1) confirm `BOOL_FIELD` round-trips `None` cleanly through default_specs generation / Hydra structured-config validation, and (2) make sure the tao-core counterpart (NVIDIA-TAO/tao-core#36) mirrors exactly this schema — same `mode` valid_options, same Optional `enabled`, same `train_logit_calibration` default — so the canonical schema and this backend copy can't drift.

nvidia_tao_pytorch/multimodal/clip/model/pl_clip_model.py
```diff
@@ -598,7 +598,6 @@ def training_step(self, batch):
             "train/lr", current_text_lr,
             on_step=True, on_epoch=False, prog_bar=True
         )
-
```
vpraveen-nv, line 601: This file's only change is deleting two blank lines — please drop it from the PR to keep the changeset focused.

vpraveen-nv (changes requested):
Thanks Chris — the per-tower mode design is a clear improvement over the enabled booleans, and the preflight-before-mutation pattern plus the new tiny-model test suite are exactly right. A few items below before I approve; the video_clip one is the only blocker.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
yangchris11: Fixed in `3db8e4a3`.

The shared LoRA utility now uses `named_logit_parameters()` when available and falls back to the existing getter/direct-attribute behavior for models such as VideoCLIP. The resolved parameters are reused for both enabling logit calibration and calculating `logit_trainable_params`.

I also strengthened the VideoCLIP regression test to verify that `logit_scale` and `logit_bias` remain trainable and that the reported parameter count is correct.

yangchris11: Validation:
  - Full VideoCLIP unit suite: 93 passed, 1 skipped
  - CLIP LoRA tests: 74 passed

nvidia_tao_pytorch/multimodal/clip/model/lora.py
yangchris11: Fixed in `2c3c5264`.

 SigLIP2 now enables gradient checkpointing through the Hugging Face root model with `use_reentrant=False`. The root model propagates this setting to both the vision and text encoders; the individual tower modules do not expose the checkpointing API directly. The existing disable path is preserved.

yangchris11: Also added regression coverage that:
  - verifies both SigLIP2 tower encoders are checkpoint-enabled with the non-reentrant setting;
  - verifies checkpointing can still be disabled;
  - freezes a tiny real `Siglip2TextModel`, injects LoRA, and runs forward/backward;
  - confirms every trainable LoRA parameter receives a gradient and at least one gradient is nonzero.

yangchris11: https://github.com/NVIDIA-TAO/tao-pytorch/blob/2c3c52644202cd59d434aba0f06ad01d62aef293/nvidia_tao_pytorch/multimodal/clip/model/adapters/siglip2.py#L170-L178

nvidia_tao_pytorch/multimodal/clip/model/lora.py
yangchris11: Thanks @nnagrajrao, this is a good catch!

Fixed in `06847008`.

`LoRALinear.scaling` is now a persistent scalar buffer, so new checkpoints carry the scale used during training. When evaluation or export reconstructs LoRA with the same rank but a different alpha, checkpoint loading restores the saved scale before `forward()` or `merge_lora()` runs.

For compatibility with pre-fix internal checkpoints that contain no scaling keys, loading falls back to the scale derived from the current PEFT `rank` and `alpha` and emits one warning requiring those values to match the original training configuration. A partially missing scaling layout is still rejected as checkpoint corruption, and
different ranks continue to fail through LoRA tensor-shape validation.

yangchris11: An example 
  ```text
  New checkpoint
  --------------
  Training config: rank=64, alpha=128  → scaling=2
  Export config:   rank=64, alpha=64   → initially constructs scaling=1
  Checkpoint load: saved scaling=2     → restores scaling=2
  Merge/export therefore uses the training-time scale.

  Legacy checkpoint 
  ---------------------------------
  Checkpoint: no persisted scaling
  Original config: rank=64, alpha=128  → fallback scaling=2
  Result: loads successfully and emits a compatibility warning.
  ```

  For legacy checkpoints, the original alpha cannot be recovered from the weights alone. The supplied configuration must therefore match the training configuration. New checkpoints no longer have this limitation because their scaling is persisted.

yangchris11: Regression coverage verifies scaling persistence, same-rank/different-alpha restoration, matching forward and merged outputs, legacy loading with a warning, and rank mismatch rejection.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
yangchris11: I agree/propose retaining the hard error. Clamping an oversized value would silently convert a likely backbone/specification mismatch into full-tower adaptation, potentially changing training cost and behavior. `num_last_blocks == depth` remains valid, and `0` remains the explicit all-block setting.

I will also update the release note to state that `num_last_blocks > depth` now fails and must be reduced or set to `0`.

yangchris11: I audited both the normal CLIP and VideoCLIP shipped specifications.

  Normal CLIP:

  - `multimodal/clip/experiment_specs/experiment_spec.yaml` is the only shipped normal CLIP specification.
  - It contains no `peft` section and no `num_last_blocks` value, so it does not rely on the previous oversized-value behavior.
  - The schema default is `num_last_blocks: 3`, but PEFT is disabled by default.

  VideoCLIP:

  - `multimodal/video_clip/experiment_specs/experiment_spec_lora.yaml` merges successfully with the structured schema and selects `internvideo2-clip-l14`.
  - Its vision tower uses `num_last_blocks: 3`; the shipped L14 architecture defines 24 vision blocks.
  - Its text tower also uses `3`, but text LoRA is disabled. All bundled MobileCLIP text configurations contain at least 4 transformer blocks.
  - The commented LoRA example in the base VideoCLIP specification uses the same value.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
yangchris11: Fixed in `1f621823`.

I checked the production call paths: CLIP and VideoCLIP each inject LoRA once during construction, while resume, evaluation, and export construct a fresh model before loading checkpoint state. There is no production path that requires re-injection.

`inject_lora()` now rejects a model that already contains `LoRALinear` adapters before resolving the new configuration, freezing parameters, or modifying modules. This prevents both inconsistent second-call statistics and silently ignored rank/alpha changes.

yangchris11: I also removed the now-unreachable existing-adapter re-enable branch and its helper, simplifying the implementation.

yangchris11: https://github.com/NVIDIA-TAO/tao-pytorch/blob/1f6218235afcd06c45c1b6237c63ec775ce3147c/nvidia_tao_pytorch/multimodal/clip/model/lora.py#L184-L191

nvidia_tao_pytorch/multimodal/clip/model/pl_clip_model.py
yangchris11: Fixed in `5601f556`

nvidia_tao_pytorch/multimodal/clip/model/lora.py
yangchris11: Fixed in `5601f556`

nvidia_tao_pytorch/multimodal/clip/model/lora.py
yangchris11: Fixed in `5601f556`.

The legacy missing-value branch has been removed. Tower `mode` is now explicit and accepts only `frozen`, `full`, or `lora`; missing or unsupported values fail with a clear `ValueError`.

This removes the dependency on OmegaConf’s serialized `"???"` representation and behaves consistently for structured configurations, dataclasses, and dictionaries.

nvidia_tao_pytorch/multimodal/clip/model/lora.py
yangchris11: Fixed in `5601f556`.

I followed the suggested mode-only approach:

- Removed the tower-level `enabled` field.
- Removed the `mode: legacy` sentinel and conflict-validation path.
- Made `mode` the sole per-tower interface, with `frozen`, `full`, and `lora` as the valid values and `frozen` as the default.
- Migrated `VideoCLIPLoRATargetConfig`, the shipped VideoCLIP LoRA specification, its commented example, and the associated tests to explicit modes.

I did not retain a compatibility layer because this interface has not appeared in a released TAO version.

CLIP schema:
  https://github.com/NVIDIA-TAO/tao-pytorch/blob/5601f5564f0c941332ca865f5b2aac078bc73c6d/nvidia_tao_pytorch/config/clip/default_config.py#L113-L139

nvidia_tao_pytorch/config/clip/default_config.py
yangchris11: Fixed across both repositories.

  tao-pytorch `5601f556` and tao-core `41dffc64` now expose the same CLIP LoRA tower schema.

nnagrajrao (approved):

vpraveen-nv (approved):
Approved. LGTM.

PR #109, opened 2026-08-22
[TAO-1796][Bugfix] video_clip: avoid PyAV frame-thread deadlock

lianqiann:
## What
Fix `video_clip` training stalls caused by PyAV frame threading during sampled decoding.

## Problem
`_load_with_pyav` used `thread_type = "AUTO"`, enabling frame threading even though sampled decoding exits before EOF. A blocked loader worker can starve its rank and eventually appear as an NCCL training hang.

## Solution
Use `thread_type = "SLICE"` to retain within-frame parallelism without requiring cross-frame decoder draining. Extend the loader regression test to assert the selected threading mode.

## Testing
- Focused loader tests: 17 passed, 1 skipped
- Full `video_clip` unit suite: 90 passed, 4 skipped
- Mixed H.264/MJPEG stress test: 240/240 decodes completed
- 1×8 A100 proof: two 1,071-step epochs completed without decoder deadlock

## Checklist
- [x] Bug resolved
- [x] Regression test added
- [x] No unrelated net changes

### DIFF
--- a/nvidia_tao_pytorch/multimodal/video_clip/dataloader/video_decode.py
+++ b/nvidia_tao_pytorch/multimodal/video_clip/dataloader/video_decode.py
@@ -196,7 +196,9 @@ def _load_with_pyav(video_path, num_frames, start_time_sec, end_time_sec,
 
     with av.open(str(video_path)) as container:
         stream = container.streams.video[0]
-        stream.thread_type = "AUTO"
+        # Frame threading can hang when sampled decoding exits before EOF;
+        # keep decoding parallel within each frame instead.
+        stream.thread_type = "SLICE"
         rate = _pyav_stream_rate(stream)
         actual_frames = _pyav_stream_length(stream, rate)
         if actual_frames <= 0:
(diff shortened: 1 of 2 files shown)


nvidia_tao_pytorch/multimodal/video_clip/dataloader/video_decode.py
```diff
@@ -196,7 +196,9 @@ def _load_with_pyav(video_path, num_frames, start_time_sec, end_time_sec,
 
     with av.open(str(video_path)) as container:
         stream = container.streams.video[0]
-        stream.thread_type = "AUTO"
+        # Frame threading can hang when sampled decoding exits before EOF;
+        # keep decoding parallel within each frame instead.
+        stream.thread_type = "SLICE"
```
vpraveen-nv, line 201: Should we add this as a configurable flag to choose? If so, can we provide this between SLICE and AUTO? or will slice will run always?

vpraveen-nv (approved):
mostly LGTM. Approved otherwise. Also, should this make release/7.2.0

nvidia_tao_pytorch/multimodal/video_clip/dataloader/video_decode.py
lianqiann: @vpraveen-nv  I recommend using SLICE unconditionally for this loader. It stops after collecting the sampled frames instead of decoding the entire video, and AUTO can enable frame threading, which may hang when decoding stops early. SLICE is also PyAV’s default and still allows parallel decoding within each frame.

PR #112, opened 2026-08-25
[TLT-5952][Bugfix] Use eval_margin for in-training Visual ChangeNet classify metrics

vpraveen-nv:

`ChangeNetPlModel.__init__` built its in-training metrics as bare
`AOIMetrics()` / `AOIMetrics()`, which falls back to the euclidean-distance
default `margin=2.0`. This PR resolves that threshold from
`model.classify.eval_margin` instead, via a new `_resolve_metric_margin()`
helper, so in-training validation uses the same operating point as the
standalone `evaluate` action.

```python
metric_margin = self._resolve_metric_margin()
self.train_metrics = AOIMetrics(metric_margin)
self.val_metrics = AOIMetrics(metric_margin)
```

`_resolve_metric_margin()` also guards the degenerate pairing: when
`difference_module='learnable'` is combined with an `eval_margin > 1.0`, the
threshold can never be exceeded, so it emits a `UserWarning` and falls back to
`0.5`. `eval_margin` is left at its `2.0` default — changing the default would
alter euclidean-mode behavior — and the `eval_margin` field description in both
config definitions is updated to say what the value actually means for each
difference module.

For `difference_module: learnable` the comparison score is
`softmax(output_diff)[:, 1]`, a probability in `[0, 1]`. `AOIMetrics.update()`
predicts FAIL only when `score > margin`, so at `margin=2.0` no sample can ever
be predicted FAIL: every sample is scored PASS. The consequences are that
`val_acc` collapses to the validation split's PASS ratio and `val_fpr` is
pinned at `0.0`, both constant for the entire run.

That makes `monitor: val_acc` / `val_fpr` checkpoint selection a no-op, breaks
any AutoML or downstream ranking that reads those metrics, and means
in-training validation and standalone `evaluate` report different numbers under
the same metric names for the same checkpoint and the same CSV. The test path
already did this correctly — `on_test_epoch_start` has always built
`AOIMetrics(model.classify.eval_margin)` — so this aligns training with
evaluation rather than introducing a new convention.

### Related issues

N/A (tracked internally as NVBug 6662965).

Yes — this is a behavior change for `difference_module: learnable` runs.

Before, on a validation split that is 98.6% PASS:

```
val_acc = 98.60, val_fpr = 0.00   # every epoch, regardless of the model
```

After, at the same checkpoint and `eval_margin: 0.3`:

```
val_acc = 1.40,  val_fpr = 100.00   # matches `visual_changenet evaluate` exactly
```

The new numbers are the correct ones, but they are real metrics rather than
constants, so any workflow that ranks or selects checkpoints on `val_acc` /
`val_fpr` may now pick a different checkpoint than it did before. That is the
intent of the fix, and it is worth knowing before rerunning existing pipelines.

Euclidean mode is unchanged unless a user had configured `eval_margin` to
something other than the previously hardcoded `2.0`; in that case in-training
metrics now honor the threshold that was already configured for evaluation.

Deliberately untouched:

- `_compute_val_far()` / `val_far` — threshold-free by construction and
  therefore never affected by this bug.
- `on_test_epoch_start` and the rest of the `evaluate` path — already correct.
- The `eval_margin` default of `2.0` and all euclidean-mode defaults.

Three unit tests added to
`tests/cv_unit_test/visual_changenet/test_classification_model_contracts.py`:

- `test_training_metrics_use_eval_margin` — train/val metrics threshold at
  `eval_margin` for both difference modules.
- `test_learnable_degenerate_margin_falls_back` — `learnable` + `eval_margin >
  1.0` warns and falls back to `0.5`.
- `test_val_metrics_discriminate_on_probability_scores` — regression guard that
  probability-valued scores actually move `total_accuracy` and `false_alarm`
  off the degenerate values.

```
$ pytest -q tests/cv_unit_test/visual_changenet/test_classification_model_contracts.py \
           tests/cv_unit_test/optical_inspection/test_metrics.py
12 passed
```

All three new tests fail against the unpatched module (3 failed, 5 passed), and
the pre-existing contract and `AOIMetrics` tests are unaffected.

Functionally verified end to end on a learnable AOI classify run: a 2-epoch
training job now reports `val_acc` / `val_fpr` that are no longer equal to the
split PASS ratio and `0.0`, and `visual_changenet evaluate` on the same
checkpoint and the same CSV at `eval_margin: 0.3` reproduces them exactly
(`Total Accuracy 1.4`, `False Alarm 100.0`). Running `evaluate` at
`eval_margin: 2.0` reproduces the old degenerate pair (`Total Accuracy 98.6`,
`False Alarm 0.0`), confirming the root cause. The `UserWarning` was confirmed
to fire on the `learnable` + `2.0` pairing.

`pre-commit` (license header, pylint, pydocstyle, flake8, secret scan) passes on
all four changed files.

Generated-by: Claude Code (Opus 5)

### Release note

```release-note
Fixed in-training val_acc and val_fpr for Visual ChangeNet classification, which were constant, meaningless values for difference_module: learnable; they are now thresholded at model.classify.eval_margin and agree with the standalone evaluate action.
```

### Checklist

- [x] My commits are signed off per the DCO (`git commit -s`) — see `CONTRIBUTING.md`
- [x] The code follows the project's style, and lint and format checks pass locally
- [x] All tests pass locally
- [x] I added or updated documentation (README, docstrings, docs pages, examples)
- [ ] I updated the version, changelog, or migration notes if this change requires it
- [ ] If this touches components that are optional to install, the imports are guarded
      so the package still works without them (reviewers: please verify this)
- [x] No secrets, credentials, proprietary data, or customer data are included in this PR
- [x] I understand this contribution is licensed under the repository's license, and that
      my commit author name and email become permanently public once merged

### Notes for reviewers

The one judgement call worth scrutinizing is the `0.5` fallback for `learnable`
+ `eval_margin > 1.0`. The alternative is to raise, but that would break every
existing spec that leaves `eval_margin` at its `2.0` default — including specs
whose runs are otherwise fine because they only monitor `val_loss` or
`val_far`. A warned fallback keeps those runs working while making the
misconfiguration visible.

Deliberately out of scope: changing the `eval_margin` default to something
inside `(0, 1)`. That would silently move the euclidean-mode operating point,
which is a separate decision from fixing the learnable path.

### DIFF
--- a/nvidia_tao_pytorch/config/visual_changenet/default_config.py
+++ b/nvidia_tao_pytorch/config/visual_changenet/default_config.py
@@ -166,7 +166,7 @@ class CNModelClassifyConfig:
         default_value=2.0,
         valid_min=0,
         valid_max="inf",
-        description="Evaluation threshold score for contrastive loss",
+        description="Score threshold separating PASS from FAIL in evaluation and in-training validation metrics. Euclidean distance for difference_module=euclidean; softmax probability in (0, 1) for difference_module=learnable.",
         automl_enabled="TRUE"
     )
     embedding_vectors: int = INT_FIELD(
--- a/nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
+++ b/nvidia_tao_pytorch/cv/visual_changenet/classification/models/cn_pl_model.py
@@ -19,6 +19,7 @@
 
 import logging
 import os
+import warnings
 import pytorch_lightning as pl
 import torch
 from torch.optim import lr_scheduler
@@ -72,8 +73,9 @@ class ChangeNetPlModel(TAOLightningModule):
         self._build_criterion()
 
         self.tensorboard = experiment_spec.train.tensorboard
-        self.train_metrics = AOIMetrics()
-        self.val_metrics = AOIMetrics()
+        metric_margin = self._resolve_metric_margin()
+        self.train_metrics = AOIMetrics(metric_margin)
+        self.val_metrics = AOIMetrics(metric_margin)
         self.dm = dm
 
         self.checkpoint_filename = 'changenet_model_classify'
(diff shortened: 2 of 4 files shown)


nnagrajrao (approved):
lgtm

PR #113, opened 2026-08-26
fix(video-clip): harden export and PEFT reporting

ramanathan831:
## Summary

- report trainable parameters after LoRA injection instead of before it
- allow VideoCLIP default-spec rendering without a mandatory results directory
- reject export without a trained checkpoint and clean up artifacts created by a failed export attempt
- serialize the in-memory OpenCLIP tokenizer so offline export does not require Hugging Face access

## NVBugs

6670447, 6670205, 6670035, 6670031, and 6670019.

## Validation

- focused default-spec regression passed
- PEFT ordering, null-checkpoint, transactional-cleanup, and offline-tokenizer boundary reproducers passed after failing on the baseline
- affected Python modules compile successfully
- the full VideoCLIP module suite was not runnable in the local environment because Lightning is not installed

### DIFF
--- a/nvidia_tao_pytorch/multimodal/video_clip/model/pl_video_clip_model.py
+++ b/nvidia_tao_pytorch/multimodal/video_clip/model/pl_video_clip_model.py
@@ -121,6 +121,12 @@ class VideoCLIPPlModel(TAOLightningModule):
         self.lora_stats = None
         if self.peft_enabled:
             self.lora_stats = inject_lora(self.model, peft_cfg)
+            trainable = sum(p.numel() for p in self.model.parameters() if p.requires_grad)
+            total = sum(p.numel() for p in self.model.parameters())
+            logging.info(
+                "LoRA trainable parameter summary: %s trainable / %s total (%.4f%%)",
+                f"{trainable:,}", f"{total:,}", 100.0 * trainable / max(total, 1),
+            )
 
         # Check if retrieval validation is configured (video_text only).
         val_cfg = getattr(self.experiment_spec.dataset, 'val', None)
--- a/nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
+++ b/nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
@@ -33,6 +33,7 @@ Functions:
     load_tokenizer: Load tokenizer from disk
 """
 
+import json
 import os
 from typing import List, Optional
 
@@ -193,7 +194,7 @@ def save_tokenizer(
     """
     os.makedirs(output_dir, exist_ok=True)
 
-    inner_tokenizer = tokenizer._tokenizer
+    inner_tokenizer = getattr(tokenizer, '_tokenizer', tokenizer)
 
     if isinstance(inner_tokenizer, SigLIP2WrappedTokenizer):
         # SigLIP2: save the HuggingFace processor's tokenizer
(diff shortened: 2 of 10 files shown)


vpraveen-nv (approved):
LGTM. approving. Please coordinate with @lianqiann on this.

nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
```diff
@@ -231,13 +232,47 @@ def save_tokenizer(
...
+            with open(os.path.join(output_dir, 'merges.txt'), 'w', encoding='utf-8') as stream:
+                stream.write('#version: 0.2\n')
+                stream.writelines(f"{left} {right}\n" for left, right in ranked_merges)
+            tokenizer_config = {
+                'tokenizer_class': 'CLIPTokenizer',
+                'model_max_length': ctx_len,
+                'bos_token': '<|startoftext|>',
+                'eos_token': '<|endoftext|>',
+                'unk_token': '<|endoftext|>',
+                'pad_token': '<|endoftext|>',
```
vpraveen-nv, line 253: Parity defect: OpenCLIP pads with token ID 0 ("!"), but this config sets `pad_token: <|endoftext|>` (ID 49407). Padded batches through the reloaded tokenizer will produce different input_ids than the training-time OpenCLIP tokenizer, which shifts text embeddings for any batch with padding. #114's version of this fix aliases `<pad>` to ID 0 precisely to keep numeric parity (and carries a round-trip test asserting exact token-ID equality). If this PR rebases on #114 this hunk drops out; if not, please align the pad ID and add a parity test.

nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
```diff
@@ -1071,6 +1075,26 @@ def run_export(experiment_config: ExperimentConfig) -> None:
...
+    model_path = export_config.checkpoint
+    if not model_path:
+        raise ValueError("A trained checkpoint is required for export")
+    output_file = export_config.onnx_file or f"{os.path.splitext(model_path)[0]}.onnx"
+    base_name = os.path.splitext(os.path.realpath(output_file))[0]
+    before = set(glob.glob(f"{base_name}*"))
+    try:
+        _run_export_impl(experiment_config)
+    except Exception:
+        for artifact in set(glob.glob(f"{base_name}*")) - before:
```
vpraveen-nv, line 1090: Two points on the cleanup wrapper (which is the right idea and completes NVBug 6670031):

1. The `{base_name}*` glob deletes *anything* new sharing the prefix — a user file like `model.notes.txt` created mid-export would be swept. Safer to enumerate the expected outputs (combined/vision/text ONNX, `*_config.yaml`, `*_tokenizer/`) and delete only those.
2. The null-checkpoint guard is duplicated here and in `_run_export_impl` — keep one (the impl's, so direct callers are also covered).

Also consider `except BaseException` rather than `Exception` so Ctrl-C mid-export cleans up too — an interrupted export is exactly the bug's repro.

nvidia_tao_pytorch/multimodal/video_clip/model/pl_video_clip_model.py
```diff
@@ -121,6 +121,12 @@ def __init__(self, experiment_spec, export=False):
         self.lora_stats = None
         if self.peft_enabled:
             self.lora_stats = inject_lora(self.model, peft_cfg)
+            trainable = sum(p.numel() for p in self.model.parameters() if p.requires_grad)
+            total = sum(p.numel() for p in self.model.parameters())
+            logging.info(
+                "LoRA trainable parameter summary: %s trainable / %s total (%.4f%%)",
```
vpraveen-nv, line 127: Minor: `inject_lora` already returns this accounting in `self.lora_stats` (`total_params`, `trainable_params`, plus `lora_params`, which this summary omits). Logging from `lora_stats` instead of recomputing keeps one source of truth and gives the LoRA-only count for free.

vpraveen-nv (approved):
minor nit reviews.

PR #114, opened 2026-08-26
[TAO-1796][Bugfix] video_clip: harden dependencies and export

lianqiann:
## What

Fix the Video-CLIP FC dependencies and five related Python workflows:

- Propagate PyAV and ONNXScript into the FC/release image.
- Report Hugging Face weight download failures clearly.
- Correct PEFT trainable-parameter reporting.
- Print default specs without requiring an output directory.
- Validate export inputs before loading the model.
- Save the OpenCLIP tokenizer for offline export.

## Problem

The release build still used the previous FC base, so the resulting image could omit PyAV and use the wrong ONNXScript version. PyAV was also present in two installation paths.

Several Video-CLIP error and export paths produced misleading output or failed too late.

## Solution

- Pin the corrected FC base digests in the development and release builds.
- Preserve the source-built `av==17.1.0` installation and remove its duplicate pip installation path.
- Assert `av==17.1.0`, `onnxscript==0.7.1`, and the absence of Decord in the release image.
- Add focused error reporting, parameter accounting, validation, and offline-tokenizer fixes.

## Testing

- Source tree: 60 passed, 1 expected skip.
- Installed release wheel: 60 passed, 1 expected skip.
- Verified all seven changed Python files in the wheel match branch HEAD.
- Verified real VP9 WebM decoding through TAO's PyAV backend.
- Verified no bundled `av.libs` and no Decord installation.
- Verified all 121 FC-base layers are preserved in the AMD64 release.
- Pylint: 10.00/10.
- Pydocstyle, flake8, license, DCO, and diff checks passed.

Validated release candidate:

`nvcr.io/nvstaging/tao/tao-toolkit-pyt@sha256:faeb58559e1d87afd16453580999c178feb345f9dc87e35b8f40098e9604dd09`

The earlier `py3-05` candidate is superseded because it lacks PyAV.

## Review follow-up

- Export intentionally requires a trained TAO checkpoint; checkpoint-less Hugging Face export is not a supported Video-CLIP workflow. Export, evaluate, and inference now expose mandatory checkpoint schema fields; the open companion NVIDIA-TAO/tao-core#30 still needs the same schema update before it merges.
- Failed exports now remove only newly created, explicitly expected artifacts, so retries do not require manual cleanup and unrelated same-prefix files remain untouched.
- OpenCLIP token ID 0 intentionally decodes as `<pad>` in the saved Hugging Face tokenizer while encoding parity is preserved.
- Independent real-checkpoint validation compared separate PyTorch and ONNX vision/text outputs on two random videos and two tokenized prompts: all outputs passed at `rtol=1e-3`, `atol=1e-3`; worst absolute error `4.62e-05`, minimum embedding cosine `0.9999999587`.

## Tracking

- NVBug 6664626
- NVBug 6664694
- NVBug 6670447
- NVBug 6670205
- NVBug 6670035
- NVBug 6670031
- NVBug 6670019
- Related: #94
- Overlaps #113 for the Python fixes; this branch additionally includes the FC dependency and HF failure fixes.

## Checklist

- [x] Bugs resolved
- [x] Regression tests added and passing
- [x] AMD64 release image validated
- [x] No Skill Bank changes

### DIFF
--- a/nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
+++ b/nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
@@ -203,6 +203,9 @@ def _save_openclip_tokenizer(tokenizer, output_dir):
     vocab = dict(tokenizer.encoder)
     # Alias OpenCLIP's padding ID without turning its real token ("!") into a
     # special token, which would change tokenization of ordinary punctuation.
+    # The duplicate ID intentionally makes the Hugging Face decoder map 0 to
+    # <pad>, not "!"; encoding fidelity and explicit padding take precedence
+    # over decode fidelity for this otherwise ambiguous ID.
     vocab["<pad>"] = 0
     with open(vocab_path, "w", encoding="utf-8") as vocab_file:
         json.dump(vocab, vocab_file, ensure_ascii=False)
--- a/nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
+++ b/nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
@@ -33,10 +33,11 @@ Functions:
     load_tokenizer: Load tokenizer from disk
 """
 
+import json
 import os
 from typing import List, Optional
 
-from transformers import AutoTokenizer
+from transformers import AutoTokenizer, CLIPTokenizer
 
 from nvidia_tao_pytorch.core.tlt_logging import logging
 from nvidia_tao_pytorch.cv.backbone_v2.text_utils import canonicalize_text
@@ -170,6 +171,61 @@ class OpenCLIPWrappedTokenizer:
         return {'input_ids': result}
 
 
+def _find_openclip_tokenizer(tokenizer):
+    """Find a wrapped OpenCLIP tokenizer with serializable BPE assets."""
+    visited = set()
+    while tokenizer is not None and id(tokenizer) not in visited:
+        visited.add(id(tokenizer))
+        if all(
+            hasattr(tokenizer, attribute)
+            for attribute in (
+                "encoder",
+                "decoder",
+                "bpe_ranks",
+                "context_length",
+                "sot_token_id",
+                "eot_token_id",
+            )
+        ):
+            return tokenizer
+        tokenizer = getattr(
+            tokenizer,
+            "_tokenizer",
+            getattr(tokenizer, "tokenizer", None),
+        )
+    return None
+
+
+def _save_openclip_tokenizer(tokenizer, output_dir):
+    """Save OpenCLIP BPE assets as an equivalent Hugging Face tokenizer."""
+    vocab_path = os.path.join(output_dir, "vocab.json")
+    merges_path = os.path.join(output_dir, "merges.txt")
+    vocab = dict(tokenizer.encoder)
+    # Alias OpenCLIP's padding ID without turning its real token ("!") into a
+    # special token, which would change tokenization of ordinary punctuation.
+    vocab["<pad>"] = 0
+    with open(vocab_path, "w", encoding="utf-8") as vocab_file:
+        json.dump(vocab, vocab_file, ensure_ascii=False)
+    with open(merges_path, "w", encoding="utf-8") as merges_file:
+        merges_file.write("#version: 0.2\n")
+        for first, second in sorted(
+            tokenizer.bpe_ranks,
+            key=tokenizer.bpe_ranks.get,
+        ):
+            merges_file.write(f"{first} {second}\n")
+
+    hf_tokenizer = CLIPTokenizer(
+        vocab_file=vocab_path,
+        merges_file=merges_path,
+        bos_token=tokenizer.decoder[tokenizer.sot_token_id],
+        eos_token=tokenizer.decoder[tokenizer.eot_token_id],
+        unk_token=tokenizer.decoder[tokenizer.eot_token_id],
+        pad_token="<pad>",
+        model_max_length=tokenizer.context_length,
+    )
+    hf_tokenizer.save_pretrained(output_dir)
+
+
 def save_tokenizer(
     tokenizer: CLIPCompatibleTokenizer,
     output_dir: str,
@@ -206,6 +262,12 @@ def save_tokenizer(
         hf_tokenizer.model_max_length = inner_tokenizer._max_length  # e.g. 64 for SigLIP2
         hf_tokenizer.save_pretrained(output_dir)
         logging.info("Saved SigLIP2 tokenizer to %s", output_dir)
+    elif openclip_tokenizer := _find_openclip_tokenizer(tokenizer):
+        _save_openclip_tokenizer(openclip_tokenizer, output_dir)
+        logging.info(
+            "Saved OpenCLIP tokenizer from loaded BPE assets to %s",
+            output_dir,
+        )
     else:
         # OpenCLIP-based (RADIO CLIP, OpenCLIP): save equivalent HuggingFace tokenizer
         from nvidia_tao_pytorch.multimodal.video_clip.utils.model_configs import (
--- a/nvidia_tao_pytorch/multimodal/video_clip/model/video_clip.py
+++ b/nvidia_tao_pytorch/multimodal/video_clip/model/video_clip.py
@@ -113,6 +113,7 @@ def build_model(experiment_config,
                 model_cfg=experiment_config.model,
                 logit_scale_init=init_logit_scale,
                 logit_bias_init=init_logit_bias,
+                log_parameters=not experiment_config.peft.enabled,
             ))
     elif model_type in radio_model_configs:
         adaptor_name = getattr(
--- a/nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
+++ b/nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
@@ -898,8 +898,15 @@ def run_export(experiment_config: ExperimentConfig) -> None:
     experiment_config : ExperimentConfig
         Experiment configuration containing export settings.
     """
-    register_checkpoint_safe_globals()
     export_config = experiment_config.export
+    model_path = export_config.checkpoint
+    if not model_path:
+        raise ValueError(
+            "A trained checkpoint is required for export. Set "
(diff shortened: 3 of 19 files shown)


nvidia_tao_pytorch/multimodal/video_clip/model/video_clip.py
```diff
@@ -113,6 +113,7 @@ def build_model(experiment_config,
                 model_cfg=experiment_config.model,
                 logit_scale_init=init_logit_scale,
                 logit_bias_init=init_logit_bias,
+                log_parameters=not experiment_config.peft.enabled,
```
vpraveen-nv: NVBug 6670447: this removes the misleading pre-injection report, but nothing replaces it. `inject_lora()` already returns the correct accounting (`total_params`, `trainable_params`, `lora_params`) and `pl_video_clip_model.py:123` stores it in `self.lora_stats` without ever logging it. The bug's expectation is that a LoRA run reports the actual 221,186 trainable params — as written, PEFT runs now report nothing. Suggest logging `lora_stats` right after the `inject_lora` call; otherwise 6670447 is only half-fixed.

nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
```diff
@@ -898,16 +898,22 @@ def run_export(experiment_config: ExperimentConfig) -> None:
     experiment_config : ExperimentConfig
         Experiment configuration containing export settings.
     """
-    register_checkpoint_safe_globals()
     export_config = experiment_config.export
+    model_path = export_config.checkpoint
+    if not model_path:
+        raise ValueError(
+            "A trained checkpoint is required for export. Set "
```
vpraveen-nv: Agree with aligning export with evaluate/inference (NVBug 6670035), but note this also deletes a capability: exporting from HF pretrained weights without a checkpoint used to be supported (the removed else-branch). If that path was intentional for pretrained-only deployments, this is a user-facing behavior removal worth a release note; if it was never legitimate, fine as is — just confirm no notebook or skill documents checkpoint-less export.

nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
```diff
@@ -923,13 +929,26 @@ def run_export(experiment_config: ExperimentConfig) -> None:
...
-        if model_path:
-            split_name = os.path.splitext(model_path)[0]
-            output_file = f"{split_name}.onnx"
-        else:
-            raise ValueError(
-                "onnx_file must be specified when exporting without checkpoint"
+        split_name = os.path.splitext(model_path)[0]
+        output_file = f"{split_name}.onnx"
+
+    vision_file = None
```
vpraveen-nv: NVBug 6670031: moving the existence checks before model load fixes the "after redoing all the work" half of the bug, but not the root complaint — a *failed* export still leaves partial ONNX artifacts behind, and the retry now aborts *immediately* instead of late. The user still has to manually delete files. Suggest either wrapping the export calls so partial outputs are unlinked on failure, or adding an `export.overwrite` escape hatch. As written I'd keep 6670031 open (or re-scope it) rather than close it on this PR.

nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
```diff
@@ -170,6 +171,61 @@ def __call__(self, text: List[str]):
...
+
+
+def _save_openclip_tokenizer(tokenizer, output_dir):
+    """Save OpenCLIP BPE assets as an equivalent Hugging Face tokenizer."""
+    vocab_path = os.path.join(output_dir, "vocab.json")
+    merges_path = os.path.join(output_dir, "merges.txt")
+    vocab = dict(tokenizer.encoder)
+    # Alias OpenCLIP's padding ID without turning its real token ("!") into a
+    # special token, which would change tokenization of ordinary punctuation.
+    vocab["<pad>"] = 0
```
vpraveen-nv: Decode-side hazard: `CLIPTokenizer` builds its decoder by inverting the vocab dict, and since `"<pad>"` is inserted last, ID 0 will decode as `<pad>` rather than OpenCLIP's `"!"`. Encoding parity is covered by the new test, decoding is not — any consumer that round-trips (e.g., debugging token dumps) will see `!` silently become `<pad>`. Worth either a decode-parity assertion in test_tokenizers.py documenting the intended difference, or a comment here stating decode fidelity for ID 0 is knowingly sacrificed for pad aliasing.

nvidia_tao_pytorch/multimodal/video_clip/utils/internvideo2_assets.py
```diff
@@ -132,9 +132,18 @@ def _hf_download(repo_id, filename, label):
...
-        repo_id=repo_id, filename=filename, revision=revision,
-    )
+    try:
+        resolved = hf_hub_download(
+            repo_id=repo_id, filename=filename, revision=revision,
+        )
+    except (OSError, ValueError) as exc:
+        source_revision = revision or "main"
+        raise RuntimeError(
+            f"Hugging Face download failed for {label}: repo={repo_id}, "
```
vpraveen-nv: NVBug 6664694: the actionable error is a real improvement, but the bug's core finding is that the *only documented* dataset/weights source is unreachable with no working fallback — clearer failure reporting diagnoses that, it doesn't fix it. Closing 6664694 on this PR alone would leave users exactly as blocked, just better informed. Suggest keeping it open pending a reachable source or documented local-checkpoint fallback in the skill (tao-skill-bank side), and noting the split on the NVBug.

nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
vpraveen-nv, line 977: NVBug 6670031: the early existence checks fix the "after redoing all the work" half, but a *failed* export still leaves partial ONNX behind, so the retry aborts immediately and the user must delete files by hand. #113 already implements the missing half — a transactional `run_export` wrapper that snapshots `{base_name}*` before the export and deletes newly-created artifacts on failure. Suggest adopting that wrapper here (or leaving 6670031 to #113 after it rebases on this PR), with two tweaks to its version: drop the duplicated null-checkpoint check, and scope the cleanup to the explicit expected outputs (`output_file` / `vision_file` / `text_file` / tokenizer dir) rather than a `{base_name}*` glob, which could sweep up unrelated files.

nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
vpraveen-nv, line 947: Agree with aligning export with evaluate/inference (NVBug 6670035), but note this also deletes a capability: exporting from HF pretrained weights without a checkpoint used to be supported (the removed else-branch). If that path was intentional for pretrained-only deployments, this is a user-facing behavior removal worth a release note; if it was never legitimate, fine as is — just confirm no notebook or skill documents checkpoint-less export.

nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
vpraveen-nv, line 209: Decode-side hazard: `CLIPTokenizer` builds its decoder by inverting the vocab dict, and since `"<pad>"` is inserted last, ID 0 will decode as `<pad>` rather than OpenCLIP's `"!"`. Encoding parity is covered by the new test, decoding is not — any consumer that round-trips (e.g., debugging token dumps) will see `!` silently become `<pad>`. Worth either a decode-parity assertion in test_tokenizers.py documenting the intended difference, or a comment here stating decode fidelity for ID 0 is knowingly sacrificed for pad aliasing.

vpraveen-nv (approved):
mostly LGTM. Please coordinate the fixes with @ramanathan831 and merge. approving.

nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
lianqiann: Fixed in 5a6bc77. run_export now rolls back only newly created explicit artifacts, including combined/separate ONNX files, encrypted-export intermediates, external data, config, and tokenizer outputs; unrelated same-prefix files are preserved.

nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
lianqiann: By design, export remains checkpoint-backed. Clarified in 3a8b2f2 and the PR description; no Video-CLIP notebook or Skill Bank workflow documents checkpoint-less export.

nvidia_tao_pytorch/multimodal/video_clip/model/tokenizers.py
lianqiann: Addressed in 0513a5d. The ID 0 decoder mapping to <pad> is explicitly documented and asserted while encoding parity remains covered.

lianqiann (commented):
Addressed all three review comments in 5a6bc77, 3a8b2f2, and 0513a5d. Focused tests and static checks pass, and independent real-checkpoint PyTorch-vs-ONNX parity passed for both separate encoders. @ramanathan831, the review fixes are ready for coordination.

christinayyw (approved):
Approving for OSRB, as `av` was already previously approved.

nvidia_tao_pytorch/multimodal/video_clip/scripts/export.py
lianqiann: Follow-up in 10738ac: export, evaluate, and inference now use mandatory checkpoint schema fields (`str` / `MISSING`), matching their runtime validation and the shipped `???` specs. Cross-repo concern: the open companion NVIDIA-TAO/tao-core#30 still carries the old optional/Hugging Face export schema and needs the same update before it merges.

vpraveen-nv (approved):
LGTM. Covers all four fixes from #113 with independent implementations, plus the PyAV/dependency work and the two extra bugs (6664626, 6664694).

Note for the record: I pushed 120aacc3 to this branch, porting the post-injection LoRA parameter summary from #113. Without it this PR had the suppression half (`log_parameters=not experiment_config.peft.enabled`) but no replacement report, so a PEFT run logged no parameter summary at all. With that in, #113 is fully superseded and can be closed.

PR #130, opened 2026-09-10
[TAO-1796][Bugfix] release image: enforce the FC video dependency contract on any base image

lianqiann:
## What

Make `release/docker/Dockerfile` enforce the FC video dependency contract on **whatever base image the build is handed**, instead of only asserting it:

- New `release/docker/ensure_fc_video_deps.sh` (rides the existing wheel `COPY`, runs in the existing single `RUN` layer — no new filesystem layers; the release image sits near overlay2's 128-layer cap).
- Reads the exact `av==` / `onnxscript==` pins from `docker/requirements-pip.txt` (single source of truth; fails closed if a pin is not exact).
- If PyAV is missing, mismatched, or not linked against the restricted FFmpeg: **source-builds** it against `/usr/local` (`--no-binary=av`, Cython 3.1.x build constraint), exactly like `docker/Dockerfile`. If the base has no restricted FFmpeg it fails the build with an actionable message — it never installs the PyPI wheel.
- Pins `onnxscript`, removes `decord` if a base ever ships it, then verifies: exact versions, no `av.libs/`/`decord.libs/`/libx264/libx265/openh264 anywhere in site-packages, PyAV resolves `libavcodec` from `/usr/local/lib`.
- Documents the CI contract in the Dockerfile header: the `X86_DIGEST`/`ARM64_DIGEST` defaults are consumed only by local builds; CI resolves the base from `docker/manifest.json` and substitutes it via BuildKit `--build-context`.
- `tests/core/test_release_fc_video_deps.py` (6 tests): pins are exact, the script resolves them, fail-closed parsing, and the Dockerfile wiring (same COPY as the wheels, runs after the wheel install, no `pip install av==` stopgap).

## Root cause of the nightly failure (`tao-pytorch-OSS-release` #57–#59, "No module named 'av'")

#114 rebuilt the base with PyAV and moved `docker/manifest.json` and the Dockerfile `ARG *_DIGEST` to `224a1610` / `2e095fb1`. The release CI never reads those Dockerfile ARGs: `publishReleaseDockerImage` (tao-jenkins-shared-library) pulls a base, flattens it, and substitutes it for the `base-<arch>` stage with `--build-context`. Until 09/08 it flattened the shared-library `TaoConfig` pin — still the pre-PyAV `0dda18f2` / `0edf38a0` — so the shipped image was built from a base with **no `av` and `onnxscript 0.6.2`** while only the wheel was compiled against the manifest base. The #114 assertion then did its job and failed the stage.

Verified on dev-box against the registry images:

| base digest | av | onnxscript | restricted FFmpeg (`/usr/local`) |
|---|---|---|---|
| `0dda18f2…` (old CI pin) | absent | 0.6.2 | present |
| `224a1610…` (manifest / #114) | 17.1.0 (source-built) | 0.7.1 | present |

Why the local `deploy.sh --build` could not reproduce it: a plain `docker build` honours the Dockerfile's `ARG X86_DIGEST` default (`224a1610`, has PyAV). The CI path replaced the base underneath the Dockerfile.

CI side is already fixed in tao-jenkins-shared-library `5f964465` (09/08): `publishReleaseDockerImage` now prefers the manifest-resolved `RESOLVED_BASE_*` and the fallback pins were refreshed. This PR is the repo side: the release image no longer depends on which base the pipeline picks, and a future base/pin drift fails loudly with a message that names the fix instead of shipping a broken image.

## Why not the stopgap (#117 / #119: `pip install av==17.1.0 onnxscript==0.7.1`)

The PyPI `av-17.1.0-cp311-abi3-manylinux_2_28_x86_64.whl` bundles its own FFmpeg under `av.libs/`, including **`libx264-….so.165` and `libx265-….so.216`** (checked by listing the wheel). That reintroduces exactly the royalty-bearing codecs TAO-2183 / FF-4 removed from the images (the base builds a codec-disabled FFmpeg and `docker/Dockerfile` has an explicit `av.libs`/libx264 leak check). Any release image built with that patch would ship libx264/libx265 again. This PR supersedes #117 and #119; they should be closed without merging.

## Testing

Probes on dev-box (`release/docker/ensure_fc_video_deps.sh` run inside the registry base images, repo mounted read-only):

- **Pinned base `224a1610`** → no-op path: keeps PyAV 17.1.0 (linked to `/usr/local/lib/libavcodec.so.62`), keeps onnxscript 0.7.1, exit 0.
- **Stale base `0dda18f2`** (the one CI shipped from) → source-builds PyAV 17.1.0 against the restricted FFmpeg in ~60 s, upgrades onnxscript 0.6.2 → 0.7.1, `av.libs` absent, `libavcodec` resolves from `/usr/local/lib`, exit 0.
- **Negative (`FFMPEG_PREFIX=/nonexistent`)** → refuses to install anything, exit 1 with the actionable error; `av` stays absent (no wheel fallback).
- `pytest tests/core/test_release_fc_video_deps.py` → 6 passed.
- Full `release/docker/Dockerfile` build with the stale base substituted the way CI does it (`--build-context base-amd64=docker-image://…@sha256:0dda18f2…`): see the follow-up comment on this PR.

## Tracking

- NVBug 6664626 (onnxscript 0.6.2 shipped instead of 0.7.1) — the mechanism behind it.
- Follow-up to #114; supersedes #117, #119.
- CI counterpart: tao-jenkins-shared-library `5f964465`, `e2c48a41`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


release/docker/ensure_fc_video_deps.sh
```diff
@@ -0,0 +1,144 @@
...
+# Make the release image's FC video dependencies match docker/requirements-pip.txt,
+# whatever base image the release build was handed.
+#
+# Why this exists: release/docker/Dockerfile only *asserted* av/onnxscript, trusting
+# the base image to carry them. The release CI does not honour the Dockerfile's
+# X86_DIGEST/ARM64_DIGEST defaults -- it substitutes its own (flattened) base via
+# BuildKit --build-context -- so a stale base pin in CI shipped a release image with
+# no PyAV and onnxscript 0.6.2, and the assert killed the nightly
+# (ModuleNotFoundError: No module named 'av'). A local `deploy.sh --build` could not
+# reproduce it because it consumed the Dockerfile's (correct) digest.
```
vpraveen-nv, line 14: Generally, I would like more comments, but for this file, not sure if we need to mention about CI internals in the comment here. We can keep it relevant to just the script and it's functionality.

release/docker/Dockerfile
```diff
@@ -1,7 +1,13 @@
...
-# FROM nvcr.io/nvstaging/tao/tao_pytorch_base_image@sha256:23c2385c3d9232cf0e8b4d37d13fc7e9c9db8fece5c0f6475e5f7924dc6ac06a
-# Select base image based on architecture
+# Select base image based on architecture.
+#
+# The *_DIGEST defaults below are only consumed by local builds (release/docker/deploy.sh).
+# The release CI resolves the base from docker/manifest.json (resolve_base_image stage),
+# flattens it and substitutes it for the base-<arch> stage with BuildKit
+# `--build-context base-<arch>=docker-image://...`, so the digests here are never read
+# there. Keep them equal to docker/manifest.json so local and CI builds start from the
+# same base, and treat docker/manifest.json as the source of truth.
```
vpraveen-nv, line 10: This header describes how the release pipeline resolves and substitutes the base (stage name, flattening, `--build-context`). That is internal CI detail and this repo is public — let's keep pipeline mechanics out of the Dockerfile. Something like this carries the part a contributor actually needs:

```
# Select base image based on architecture.
# docker/manifest.json is the source of truth for these digests; keep them in sync.
# Builds may override the base-<arch> stage via BuildKit --build-context, so the
# image must not assume anything about the base beyond what ensure_fc_video_deps.sh verifies.
```

release/docker/ensure_fc_video_deps.sh
```diff
@@ -0,0 +1,144 @@
...
+    ffmpeg_prefix_seen="$(pkg-config --variable=prefix libavcodec)"
+    [[ "${ffmpeg_prefix_seen}" == "${FFMPEG_PREFIX}" ]] \
+        || die "pkg-config resolved libavcodec from '${ffmpeg_prefix_seen}', expected '${FFMPEG_PREFIX}'."
+    constraints="$(mktemp)"
+    printf '%s\n' "${PYAV_BUILD_CYTHON}" > "${constraints}"
+    pip uninstall -y av || true
+    pip install --no-cache-dir --force-reinstall --no-deps --no-binary=av \
+        --build-constraint "${constraints}" "${av_requirement}"
+    rm -f "${constraints}"
+fi
```
vpraveen-nv, line 118: This block (Cython build-constraint, `--no-binary=av` install, decord removal, and the forbidden-lib `find` below) duplicates `docker/Dockerfile` L124–155 almost verbatim, so two places must now agree on the codec rules and the Cython pin. Since the script already reads the pin from `docker/requirements-pip.txt`, could `docker/Dockerfile` `COPY` and run this same script (or the script live in `docker/` next to the requirements it reads)? One file owning the PyAV/FFmpeg contract for both images removes exactly the drift risk this PR is fixing. At minimum, cross-reference the two so the next Cython bump touches both.

release/docker/ensure_fc_video_deps.sh
```diff
@@ -0,0 +1,144 @@
...
+    echo "onnxscript state: installed='${current_onnxscript:-<absent>}' wanted='${onnxscript_version}' -> installing pinned version."
+    pip install --no-cache-dir --no-deps "${onnxscript_requirement}"
+fi
+
+# ---- Verify: exact versions, no decord, no bundled codec libraries, PyAV on the restricted FFmpeg ----
+bad_libs="$(find "${site_packages}" \
+    \( -path '*/av.libs/*' -o -path '*/decord.libs/*' \
+       -o -name 'libx264*.so*' -o -name 'libx265*.so*' -o -name 'libopenh264*.so*' \) -print)"
+[[ -z "${bad_libs}" ]] || die "forbidden bundled video codec libraries in ${site_packages}:"$'\n'"${bad_libs}"
+av_links_restricted_ffmpeg || die "PyAV is not linked against ${FFMPEG_PREFIX}/lib/libavcodec."
```
vpraveen-nv, line 134: On the rebuild path this runs with `LD_LIBRARY_PATH=${FFMPEG_PREFIX}/lib:...` still exported (L106), so `ldd` resolves `libavcodec` from `${FFMPEG_PREFIX}/lib` by construction — it verifies what the script asked for, not what the runtime will see (the image sets no such `ENV`). Suggest scoping the exports to the `pip install` line, or running the verification under `env -u LD_LIBRARY_PATH -u PKG_CONFIG_PATH`, so the check reflects the shipped environment (ldconfig / RUNPATH).

release/docker/ensure_fc_video_deps.sh
```diff
@@ -0,0 +1,144 @@
...
+    rm -f "${constraints}"
+fi
+
+# ---- ONNXScript: exact pin (pure Python; its deps -- onnx, numpy, ml_dtypes, typing_extensions -- ship in the base) ----
+current_onnxscript="$(installed_version onnxscript)"
+if [[ "${current_onnxscript}" == "${onnxscript_version}" ]]; then
+    echo "onnxscript ${current_onnxscript} already present -- keeping it."
+else
+    echo "onnxscript state: installed='${current_onnxscript:-<absent>}' wanted='${onnxscript_version}' -> installing pinned version."
+    pip install --no-cache-dir --no-deps "${onnxscript_requirement}"
```
vpraveen-nv, line 126: nit: `--no-deps` keeps the layer tight, but a base carrying an older `onnxscript` may also carry older `onnx`/`ml_dtypes`/`typing_extensions` than 0.7.1 wants. A `pip check`-style guard limited to onnxscript after this install would surface that at build time instead of at first export.

release/docker/ensure_fc_video_deps.sh
```diff
@@ -0,0 +1,144 @@
...
+try:
+    import av  # noqa: F401
+except Exception:
+    sys.exit(1)
+ext = glob.glob(os.path.join(os.path.dirname(av.__file__), "container", "core*.so"))
+if not ext:
+    sys.exit(1)
+out = subprocess.run(["ldd", ext[0]], capture_output=True, text=True).stdout
+line = next((l for l in out.splitlines() if "libavcodec" in l), "")
+sys.exit(0 if os.path.join(prefix, "lib") + "/" in line else 1)
```
vpraveen-nv, line 89: nit: this is a substring test on the `ldd` text — `/usr/local/lib/` also matches `/usr/local/lib/x86_64-linux-gnu/…` or `/opt/usr/local/lib/…`. Parsing the `=> <path>` token and comparing `os.path.dirname(path) == os.path.join(prefix, "lib")` is exact. Low risk, but this is the check that guards the licensing rule.

tests/core/test_release_fc_video_deps.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def test_release_dockerfile_runs_the_contract_after_installing_wheels():
+    """The dependency fix-up runs in the wheel RUN layer, after the TAO wheels are installed."""
+    dockerfile = RELEASE_DOCKERFILE.read_text()
+    install = dockerfile.index("python -m pip install '{}'")
+    ensure = dockerfile.index("ensure_fc_video_deps.sh /opt/nvidia/wheels/requirements-pip.txt")
+    assert install < ensure
+    # The stopgap (PyPI wheel, bundles libx264/libx265) must not come back.
+    assert not re.search(r"pip install ['\"]?av==", dockerfile)
```
vpraveen-nv, line 73: These text-shape tests will trip on harmless edits (splitting the `RUN` for readability, re-indenting), and `.index()` raises `ValueError` with no context. Fine as a tripwire, but please say in the docstring what rule is being guarded (single RUN layer, no PyPI `av` wheel) so whoever hits it knows what they are allowed to change, and turn the `.index()` calls into asserts with a message.

PR #132, opened 2026-09-10
[TAO-2529][Feature] Add strict launch validation for DINOv3 SSL DEFT

nnagrajrao:

Keep DINOv3 DEFT on TAO's native Lightning path while adding scheduler-owned device-mask checks, fail-closed multi-node rendezvous validation, and real topology regressions; no torchrun launcher is introduced.

**DEFT for SSL**

Within a repository, arrows are Git base dependencies and should be reviewed and merged left to right. Cross-repository links are runtime/API dependencies. This increment is based on `main`, contains one signed-off commit, and is headed by `nnagrajrao/deft/01-foundations`.

DEFT workers must honor platform allocations and reject stale or conflicting distributed state without changing launch behavior for other TAO models.

### Related issues

DINOv3 DEFT gains opt-in strict Lightning launch validation; all other models retain existing defaults.

The final clean committed snapshots were archived and tested inside the approved TAO 7.2 rc36 containers:

- Final repository tips: Skill Bank `edef2e5d`, PyTorch `c22f3b80`, Data Services `7f2497e3`, Core `b3fd05f`.
- Source CI handoff: Skill Bank validation/hooks plus 27 tests; PyTorch hooks plus 323 passed and 4 expected weight-dependent skips; Data Services hooks plus 216 passed; Core hooks plus 19 passed.
- Broader QA: PyTorch 263 passed and 4 expected skips; Data Services 211 passed; Core 19 passed; Skill Bank 27 passed.
- Cross-repository dummy-data smoke: 5 passed, covering real image materialization/consumption, a DS LocalRunner process boundary into precomputed GRIT, exact embedding binding, schema equality, and DS/PyTorch attestation parity.
- Real Lightning smoke covered two local workers and two real nodes without torchrun.
- Independent adversarial panel re-review found no remaining source, contract, isolation, or test-coverage blocker.

Approved images were pinned to PyTorch `sha256:83eb5598411bf5a2b28f366e8861153e8da77f805edef0d200baa95bf3c5dc62` and DS `sha256:ab55b92d8c0eb570c29661dc8fe4a03dfe76cd5cc3a685ed77edcfc221c0f87a`. The source-free packaged probes fail only because rc36 predates the new PyTorch/DS modules and Core schema; rebuilding compatible release images remains a release-time requirement. No Dockerfile or dependency manifest changed.

Each GitLab integration branch is one signed-off commit over current `origin/main` (`0 1` ancestry), and its tree exactly matches the final GitHub stack tip. Production-scale throughput, real-data quality, full multi-node/FSDP training, and rebuilt packaged images remain outside this validation.

### Release note

```release-note
DINOv3 DEFT validates Lightning launch state and publishes rank-safe runtime metadata.
```

### Checklist

- [ ] Published release images contain the complete feature stack — pending normal TAO image publication.

### Notes for reviewers

Review base: `main`; head: `nnagrajrao/deft/01-foundations`. This PR is ready for review and is not a draft. The scope is DINOv3 SSL DEFT only; other models keep their existing behavior.


nvidia_tao_pytorch/ssl/dinov3/entrypoint/dinov3.py
```diff
@@ -29,7 +30,12 @@ def main():
     args, unknown_args = command_line_parser(parser, subtasks)
 
     # Launch the subtask.
-    launch(vars(args), unknown_args, subtasks, network="dinov3")
+    launch(
+        vars(args), unknown_args, subtasks, network="dinov3",
+        use_torchrun=args.subtask == "train",
```
vpraveen-nv: any reason why we default to torchrun for train?

nvidia_tao_pytorch/core/entrypoint.py
```diff
@@ -264,8 +270,9 @@ def launch(args, unknown_args, subtasks, network=None):
         log_file = f"{logs_dir}/{os.getenv('JOB_ID')}/microservices_log.txt"
 
     # Create a system call.
-    if network in LIGHTNING_EXCLUDED_NETWORKS:
-        os.environ["CUDA_VISIBLE_DEVICES"] = os.environ["TAO_VISIBLE_DEVICES"]
+    if use_torchrun or network in LIGHTNING_EXCLUDED_NETWORKS:
+        if not (preserve_cuda and os.environ.get("CUDA_VISIBLE_DEVICES")):
+            os.environ["CUDA_VISIBLE_DEVICES"] = os.environ["TAO_VISIBLE_DEVICES"]
```
vpraveen-nv: What's the scope of preserve_cuda?

vpraveen-nv (commented):
minor nitpicks here.

nvidia_tao_pytorch/ssl/dinov3/entrypoint/dinov3.py
nnagrajrao: Agreed—there is no need to default DINOv3 training to torchrun. This revision removes that opt-in and keeps the existing Python-to-Lightning DDP/FSDP path. For DEFT, the Data Services adapter starts one process per node and an explicit LightningEnvironment owns local worker spawning; Slurm ownership, stale-rank rejection, and real two-worker launch regressions now cover that boundary.

nvidia_tao_pytorch/core/entrypoint.py
nnagrajrao: `preserve_cuda` was only intended to retain a scheduler-owned CUDA_VISIBLE_DEVICES mask in the proposed torchrun branch. Since DINOv3 now stays on the Python/Lightning path, that mask is inherited naturally, so the option and its shared-launcher behavior have been removed entirely. Existing RT-DETR torchrun behavior and other models remain unchanged.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
```diff
@@ -0,0 +1,23 @@
...
+
+"""Rank-owned runtime spec publication for DINOv3 training."""
+
+import os
+from pathlib import Path
+
+from omegaconf import OmegaConf
+
+
+def publish_runtime_spec(cfg, results_dir: str) -> None:
```
vpraveen-nv: 🟠 **Major**: Nothing in this PR calls `publish_runtime_spec` — the only references are the two new tests. The first production caller is `ssl/dinov3/scripts/train.py` in #133 (`@monitor_status(..., spec_writer=publish_runtime_spec)`). Same for the `spec_writer` kwarg on `monitor_status`. As an increment this ships dead code, and the description's "rank-zero atomic publication is DINOv3-owned" isn't true of this tree. Either add the one-line decorator change to `scripts/train.py` here (then the PR is self-contained and testable end to end), or move this module + the decorator hook + `tests/core/test_workflow_decorator.py` into #133.

nvidia_tao_pytorch/core/decorators/workflow.py
```diff
@@ -39,7 +39,10 @@ def _func(cfg, **kwargs):
             os.makedirs(cfg["results_dir"], exist_ok=True)
             results_dir = cfg["results_dir"]
 
-            OmegaConf.save(cfg, os.path.join(results_dir, "experiment.yaml"))
+            if spec_writer is None:
```
vpraveen-nv: ❔ **Question**: Why is the atomic write an opt-in hook rather than the default? Every Lightning DDP task re-executes the script once per local rank and all of them hit `OmegaConf.save` on the same `experiment.yaml` today — that's the actual bug, and it isn't DINOv3-specific. temp-file + `os.replace` in the default branch is strictly better with no reader-visible change; rank-gating can stay opt-in if you're worried about node-local results dirs. Adding a callback parameter to the shared decorator for one model is API surface I'd rather not carry.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
```diff
@@ -0,0 +1,23 @@
...
+    """Atomically publish one DINOv3 runtime spec from global rank zero."""
+    rank_coordinates = ("RANK", "NODE_RANK", "LOCAL_RANK")
+    if any(int(os.getenv(name, "0")) != 0 for name in rank_coordinates):
+        return
+    destination = Path(results_dir) / "experiment.yaml"
+    temporary_path = destination.with_name(".experiment.yaml.tmp")
```
vpraveen-nv: 🟡 **Minor**: Fixed temp name `.experiment.yaml.tmp` in a shared directory. Two writers that both look like rank 0 (retry launched while a stale process is alive, or two experiments pointed at the same `results_dir`) will collide: one process's `finally: unlink` can remove the other's temp file before its `replace()`, which then raises `FileNotFoundError`. Use `tempfile.mkstemp(dir=results_dir, prefix=".experiment.yaml.")` then `os.replace`.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
```diff
@@ -0,0 +1,23 @@
...
+import os
+from pathlib import Path
+
+from omegaconf import OmegaConf
+
+
+def publish_runtime_spec(cfg, results_dir: str) -> None:
+    """Atomically publish one DINOv3 runtime spec from global rank zero."""
+    rank_coordinates = ("RANK", "NODE_RANK", "LOCAL_RANK")
+    if any(int(os.getenv(name, "0")) != 0 for name in rank_coordinates):
```
vpraveen-nv: 🟡 **Minor**: `int(os.getenv(name, "0"))` raises `ValueError` if the variable exists but is empty (`export RANK=` is common in scheduler wrappers), and this runs before `monitor_status` sets up the status logger, so it surfaces as an unlogged traceback. `int(os.getenv(name) or "0")` is the safe form.

nvidia_tao_pytorch/core/entrypoint.py
```diff
@@ -246,6 +249,11 @@ def launch(args, unknown_args, subtasks, network=None):
             ]
         except Exception as e:
             logging.warning(f"[Multinode] Error overriding configs: {e}")
+            if strict_multinode:
```
vpraveen-nv: 🟠 **Major**: Two things. (a) Raising a bare `RuntimeError` here escapes `launch()` before the `process_passed`/telemetry/`sys.exit(1)` tail, so the caller gets a traceback instead of the usual `Execution status: FAIL` + exit-code path every other launch failure uses. Log and `sys.exit(1)`, or route through the same tail. (b) 'Strict launch validation' is generous for what this is: it only converts one pre-existing `except Exception` (which also swallows the spec-file rewrite failing, not just `validate_configs`) into a raise. When `WORLD_SIZE` is unset but the spec says `train.num_nodes > 1`, or `NUM_GPU_PER_NODE` disagrees with `device_count()`, nothing is validated even in strict mode. If DEFT relies on this being fail-closed those cases need covering too, or the docstring/title should say 'no single-node fallback' rather than 'strict validation'.

nvidia_tao_pytorch/core/entrypoint.py
```diff
@@ -112,13 +112,16 @@ def dual_output(log_file=None):
         yield sys.stdout, None
 
 
-def launch(args, unknown_args, subtasks, network=None):
+def launch(args, unknown_args, subtasks, network=None, *, strict_multinode=False):
```
vpraveen-nv: 🟡 **Minor**: The flag is only ever set from `TAO_STRICT_MULTINODE` in the DINOv3 entrypoint, so a generically-named env var silently does nothing for every other network. If the intent is an env-driven launcher knob, read it here inside `launch()` (one place, works for all networks, and the test on `test_visible_devices.py:36` would then actually exercise something). If it's DINOv3-only, name it that way or make it an explicit CLI flag instead of a hidden env var.

nvidia_tao_pytorch/ssl/dinov3/entrypoint/dinov3.py
```diff
@@ -29,7 +30,10 @@ def main():
     args, unknown_args = command_line_parser(parser, subtasks)
 
     # Launch the subtask.
-    launch(vars(args), unknown_args, subtasks, network="dinov3")
+    launch(
+        vars(args), unknown_args, subtasks, network="dinov3",
+        strict_multinode=os.environ.get("TAO_STRICT_MULTINODE") == "1",
```
vpraveen-nv: 🟡 **Minor**: `TAO_STRICT_MULTINODE` isn't documented anywhere (`ssl/dinov3/README.md`, `docs/`, module docstring) — I grepped the head tree; the only mentions are this line and the tests. The checklist ticks 'added or updated documentation' but no doc file is in the diff. Please document the variable, its values, and which entrypoints honour it.

nvidia_tao_pytorch/ssl/dinov3/entrypoint/dinov3.py
```diff
@@ -29,7 +30,10 @@ def main():
     args, unknown_args = command_line_parser(parser, subtasks)
 
     # Launch the subtask.
-    launch(vars(args), unknown_args, subtasks, network="dinov3")
+    launch(
```
vpraveen-nv: 🟠 **Major**: Title/commit mismatch: PR title says `[Bugfix] Add strict launch validation`, the commit says `Add opt-in launch hooks`. The diff changes no default behaviour and fixes no reported bug — it adds two opt-in hooks. This should be `[Feature]` (or `[Chore]`), and the description trimmed to what *this* increment does; most of the current body (Arrow SIGABRT, DS image compatibility, GRIT smoke, FAISS) describes other PRs in the stack and makes it hard to see what's being claimed for these 6 files.

tests/core/test_workflow_decorator.py
```diff
@@ -0,0 +1,62 @@
...
+    publish_runtime_spec(config, str(tmp_path))
+    assert not destination.exists()
+
+    monkeypatch.setenv(rank_variable, "0")
+    publish_runtime_spec(config, str(tmp_path))
+    assert OmegaConf.load(destination) == config
+    assert not list(tmp_path.glob(".experiment.*.tmp"))
+
+
+def test_default_status_writer_is_unchanged_on_nonzero_rank(tmp_path, monkeypatch):
```
vpraveen-nv: 🟡 **Minor**: This asserts that a non-zero-rank process *does* overwrite `experiment.yaml` via the default path — i.e. it enshrines the current race as a contract. If we later fix the default writer (see my `workflow.py` comment) this test has to be deleted. I'd drop it or reword it to test only that the default path still writes the file. Also `tests/core` now imports `nvidia_tao_pytorch.ssl.dinov3.utils.runtime_spec`; core tests shouldn't depend on a model package.

tests/distributed/test_visible_devices.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def popen(argv, **_kwargs):
+        calls.append(argv)
+        return SimpleNamespace(stdout=io.StringIO(), wait=lambda: None, returncode=0)
+
+    for key in ("WORLD_SIZE", "NODE_RANK", "RANK", "JOB_ID"):
+        monkeypatch.delenv(key, raising=False)
+    monkeypatch.setenv("CUDA_VISIBLE_DEVICES", "GPU-a,GPU-b")
+    # An inherited DEFT flag changes validation, not a model's launch policy.
+    monkeypatch.setenv("TAO_STRICT_MULTINODE", "1")
```
vpraveen-nv: ⚪ **Nit**: `TAO_STRICT_MULTINODE` is only read in `dinov3.main()`, not in `launch()`, so this `setenv` is a no-op and the comment above it is asserting something the test doesn't exercise. Either drive the test through `dinov3.main()` or remove the setenv and comment.

tests/distributed/test_visible_devices.py
```diff
@@ -0,0 +1,71 @@
...
+    monkeypatch.setenv("WORLD_SIZE", "2")
+    monkeypatch.setenv("TAO_STRICT_MULTINODE", "1")
+    monkeypatch.delenv("JOB_ID", raising=False)
+    monkeypatch.setattr(entrypoint, "TELEMETRY_AVAILABLE", False)
+    monkeypatch.setattr(entrypoint.torch.cuda, "device_count", lambda: 1)
+
+    def invalid(_logger):
+        raise ValueError("bad rendezvous")
+
+    monkeypatch.setattr(entrypoint, "validate_configs", invalid)
```
vpraveen-nv: 🟡 **Minor**: With `validate_configs` monkeypatched to raise, this only proves the exception plumbing (fine), but the test name reads as if validation logic were covered. Also `launch()` writes `os.environ["TAO_VISIBLE_DEVICES"]` (and `RANK` on the torchrun path) directly, so those leak into subsequent tests. Pre-seed with `monkeypatch.setenv`/`delenv` or wrap in `mock.patch.dict(os.environ)`.

vpraveen-nv (commented):
minor comments.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
nnagrajrao: Fixed in the restacked #133. The DINOv3 publisher now uses `tempfile.mkstemp` in the destination directory with a unique prefix, fsyncs the file, performs `os.replace`, fsyncs the directory, and removes only its own temporary path.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
nnagrajrao: Fixed in the restacked #133: rank parsing now uses `os.environ.get(name) or "0"`, so absent and explicitly empty scheduler exports are both treated safely as rank zero. The behavior has a regression test.

nvidia_tao_pytorch/core/decorators/workflow.py
nnagrajrao: The final design keeps the shared default unchanged so this feature cannot alter other TAO models. In #133 the callback-shaped API was replaced with a single backward-compatible `write_experiment_spec` boolean that defaults to `True`; only DINOv3 DEFT opts out and uses its rank-owned atomic publisher. Regression coverage proves existing callers still take the historical path.

nvidia_tao_pytorch/core/entrypoint.py
nnagrajrao: Redesigned rather than extending the old fallback switch. PR #132 now resolves the sealed train spec plus supported Hydra allocation overrides, validates positive node/GPU counts, contiguous local IDs, visible GPU capacity, required rendezvous fields, node-rank range, and absence of inherited worker/torchelastic state before starting a child. Validation errors are retained and raised inside the existing `launch()` try/telemetry path as parameter errors, producing the normal failure status and exit code; parameterized tests cover the fail-closed cases.

nvidia_tao_pytorch/core/entrypoint.py
nnagrajrao: Fixed. The generic flag was removed. The final contract is `TAO_REFINEMENT_LIGHTNING_LAUNCH=1`, and `launch()` honors it only for `network == "dinov3"` plus the `train` subtask. Other networks and DINOv3 actions retain their existing launch policy, which is explicitly regression-tested.

tests/distributed/test_visible_devices.py
nnagrajrao: Fixed and expanded. Tests now separately cover allocation resolution, validation logic, normal failure/telemetry plumbing, malformed specs, local GPU contracts, inherited worker keys, and real 1-node/2-worker plus 2-node Lightning topology. The launch helper pre-seeds `TAO_VISIBLE_DEVICES` through `monkeypatch`, and all scheduler/worker variables are set or deleted through `monkeypatch`, so direct launcher writes are restored at fixture teardown.

tests/core/test_workflow_decorator.py
nnagrajrao: Resolved by the restack. PR #132 no longer changes the shared workflow decorator and no `tests/core` test imports a model package. Runtime publication and its model-owned tests are in #133; PR #132 tests only launcher behavior and real Lightning worker topology.

nvidia_tao_pytorch/ssl/dinov3/entrypoint/dinov3.py
nnagrajrao: Fixed. The PR title and commit are both `[TAO-2529][Feature]` and describe DINOv3 DEFT Lightning launch validation. The description now scopes this increment, explains DEFT, and links the ordered cross-repository PR stack; Arrow/GRIT/orchestration claims are assigned to their owning increments.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
nnagrajrao: Resolved by restacking as suggested. PR #132 no longer contains the DINOv3 runtime-spec module, the status-writer hook, or their tests; those now live together in #133 with the production call from `ssl/dinov3/scripts/train.py`. PR #132 is limited to the node-level launch boundary and real Lightning topology regressions.

tests/distributed/test_visible_devices.py
nnagrajrao: Fixed by removing `TAO_STRICT_MULTINODE`. The replacement tests drive the real `TAO_REFINEMENT_LIGHTNING_LAUNCH` contract through `entrypoint.launch()`, assert plain Python for DINOv3 DEFT, and assert that existing DINOv3/NVDINOv2 Python and RT-DETR torchrun policies remain unchanged when the flag is absent.

nvidia_tao_pytorch/ssl/dinov3/entrypoint/dinov3.py
nnagrajrao: Fixed directly in PR #132. The DINOv3 README now documents the internal adapter flag, the only enabling value (`1`), the required node/rendezvous variables, the sole honoring entrypoint (`dinov3 train`), unchanged behavior when unset, and Lightning ownership of local workers without torchrun.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
vpraveen-nv: Verified at 262a796 — `runtime_spec.py`, the decorator hook and `tests/core/test_workflow_decorator.py` are out of this PR and land in #133 next to their caller (`scripts/train.py:137`). Resolved, thanks.

nvidia_tao_pytorch/core/decorators/workflow.py
vpraveen-nv: Acknowledged — keeping the shared default byte-for-byte unchanged and replacing the callback with a `write_experiment_spec` bool (in #133) is a reasonable call for this stack. I still think temp-file + `os.replace` should become the default in `workflow.py` for every Lightning DDP task, but that's a separate follow-up, not on you here. Resolving.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
vpraveen-nv: Verified in #133 (`utils/runtime_spec.py:24-42`): `tempfile.mkstemp(dir=results_dir, prefix='.experiment.yaml.')`, fsync file, `os.replace`, fsync dir, unlink only its own temp path; covered by `test_concurrent_rank_zero_writers_use_unique_atomic_files`. Resolved.

nvidia_tao_pytorch/ssl/dinov3/utils/runtime_spec.py
vpraveen-nv: Verified in #133 (`runtime_spec.py:15`): `int(os.environ.get(name) or "0")`, with `test_empty_rank_exports_publish_as_rank_zero`. Resolved.

nvidia_tao_pytorch/core/entrypoint.py
vpraveen-nv: Verified at 262a796 — both halves addressed. (a) `launch_validation_error` is raised inside the existing `try` (entrypoint.py:423-424) so it takes the `Parameter validation error` → `user_error=True` → telemetry → `Execution status: FAIL` / `sys.exit(1)` path; `test_native_dinov3_deft_invalid_allocation_uses_normal_failure_path` proves it. (b) missing `WORLD_SIZE`, `NUM_GPU_PER_NODE` ≠ `train.num_gpus`, `WORLD_SIZE` ≠ `train.num_nodes`, out-of-range `NODE_RANK`, and `num_gpus > device_count()` are all rejected now. Resolved.

nvidia_tao_pytorch/core/entrypoint.py
vpraveen-nv: Verified — `_native_dinov3_deft_requested` is evaluated once inside `launch()` and gated on `network == "dinov3"` + `subtask == "train"`; `test_existing_launch_modes_are_unchanged` covers nvdinov2/dinov3/rtdetr without the flag. Resolved.

nvidia_tao_pytorch/ssl/dinov3/entrypoint/dinov3.py
vpraveen-nv: Verified — `ssl/dinov3/README.md:469-477` documents the flag, the only enabling value, the required env vars and the sole honoring entrypoint. Resolved (one wording point on the `LightningEnvironment` sentence left as a separate comment).

nvidia_tao_pytorch/ssl/dinov3/entrypoint/dinov3.py
vpraveen-nv: Verified — title and commit are `[TAO-2529][Feature] ...` and the description is scoped to this increment with the stack map. Resolved.

tests/core/test_workflow_decorator.py
vpraveen-nv: Moot for #132 — no `tests/core` changes in this PR any more. In #133 `tests/core/test_workflow_decorator.py` imports only `core.*`; no model-package import. Resolved.

tests/distributed/test_visible_devices.py
vpraveen-nv: Verified — the flag is exercised through `entrypoint.launch()` in the new `test_native_dinov3_deft_*` tests. Resolved.

tests/distributed/test_visible_devices.py
vpraveen-nv: Verified — `_launch` pre-seeds `TAO_VISIBLE_DEVICES`, deletes `JOB_ID`, and every scheduler/worker var goes through `monkeypatch`, so `launch()`'s direct `os.environ` writes are restored at teardown. Validation logic is now tested directly rather than via a monkeypatched `validate_configs`. Resolved.

nvidia_tao_pytorch/ssl/dinov3/README.md
```diff
@@ -466,6 +466,16 @@ unchanged.
...
 
+### DINOv3 DEFT Lightning launch contract
+
+The Data Services adapter starts one ordinary TAO process per node and sets
+`TAO_REFINEMENT_LIGHTNING_LAUNCH=1` together with `WORLD_SIZE` (node count),
+`NUM_GPU_PER_NODE`, `NODE_RANK`, `MASTER_ADDR`, and `MASTER_PORT`. This internal
+adapter flag is honored only by `dinov3 train`, and only the value `1` enables
+the DEFT path; when it is absent or has any other value, existing TAO launch
+behavior is unchanged. Lightning then creates and coordinates the local workers
+through `LightningEnvironment`; DEFT does not add a `torchrun` layer.
```
vpraveen-nv, line 477: 🟡 **Minor**: This sentence isn't true of this PR's tree yet: at 262a796 `ssl/dinov3/scripts/train.py` builds the `Trainer` without `plugins=[LightningEnvironment()]` — that line arrives in #133 (`_refinement_trainer_plugins`). Without it Lightning auto-detects the cluster environment, so a DEFT node process started inside a Slurm allocation (`SLURM_NTASKS` set) or a Kubeflow pod picks `SLURMEnvironment`/`KubeflowEnvironment`, whose `creates_processes_externally=True` means *no* local workers are spawned and `NODE_RANK`/`MASTER_*` are ignored. Since the launcher-side validation and the trainer-side plugin are two halves of one contract, I'd move the 5-line `_refinement_trainer_plugins` hook + `plugins=` kwarg into this PR (then this paragraph is accurate on its own), or reword to say the plugin is wired in the training script.

nvidia_tao_pytorch/core/entrypoint.py
```diff
@@ -34,6 +34,122 @@
...
+    if declared_nodes != expected_nodes:
+        raise ValueError(f"{prefix} WORLD_SIZE differs from train.num_nodes")
+    if declared_gpus != expected_gpus:
+        raise ValueError(f"{prefix} NUM_GPU_PER_NODE differs from train.num_gpus")
+    if not 0 <= node_rank < expected_nodes:
+        raise ValueError(f"{prefix} NODE_RANK is out of range")
+    if not 1024 <= master_port <= 65535:
+        raise ValueError(f"{prefix} MASTER_PORT is invalid")
+    if expected_gpus > torch.cuda.device_count():
+        raise ValueError(f"{prefix} requests more GPUs than are visible")
```
vpraveen-nv, line 150: 🟡 **Minor**: The DEFT path now bypasses `validate_configs(logging)` entirely (line 350 is guarded with `and not native_dinov3_deft`) and re-implements its range checks here, but drops the two network checks it had: `socket.gethostbyname(MASTER_ADDR)` and the rank-0 `bind((MASTER_ADDR, MASTER_PORT))` probe. Those catch a mistyped master host / port already in use *before* a 2-node rendezvous hangs for the NCCL timeout. Suggest ending this function with `validate_configs(logging)` — its `ValueError`s already get the `Parameter validation error:` prefix from the `except` in `launch()` — instead of carrying two validators that will drift.

nvidia_tao_pytorch/core/entrypoint.py
```diff
@@ -168,10 +284,30 @@ def launch(args, unknown_args, subtasks, network=None):
     # Set gpus - overwrite if fields are inconsistent
     # Precedence for gpu setting: env > cmdline > specfile > default
```
vpraveen-nv: ⚪ **Nit**: Stale for the branch below it: on the DEFT path there is no precedence at all — env, Hydra override and spec must agree or the launch fails. Worth a one-line note so the next reader doesn't assume `WORLD_SIZE` still wins.

vpraveen-nv (commented):
Thanks for the restack — I re-read the whole PR at 262a796. This is now a different (and much better) increment: `TAO_STRICT_MULTINODE` and the `spec_writer`/runtime-spec code are gone, `launch()` reads `TAO_REFINEMENT_LIGHTNING_LAUNCH=1` once and only for `dinov3 train`, the allocation resolver/validator fails closed on node/GPU/rendezvous/inherited-worker state, and validation errors go through the existing try/telemetry tail as `Parameter validation error` (verified the `user_error=True` → `Execution status: FAIL` → exit 1 path in the new tests). The sealed spec is no longer rewritten, and the two real-process Lightning tests prove the topology rather than mocking it. All 11 earlier threads are resolved or moot — replied inline.

Fresh pass found no blockers: one cross-PR consistency point (the README's `LightningEnvironment` claim depends on the `plugins=` line that lives in #133), a suggestion to reuse `validate_configs` for the network checks the new validator dropped, and a stale comment. DCO signed, static checks clean. Happy to approve once the README/plugin point is settled.

nvidia_tao_pytorch/ssl/dinov3/README.md
nnagrajrao: Moved `_refinement_trainer_plugins` and the Trainer `plugins=` call into #132. This increment now owns both halves of the Lightning launch contract; its standalone distributed suite passes 39 tests, including scheduler-environment isolation.

Revision: `9d3c8e46`.

nvidia_tao_pytorch/core/entrypoint.py
nnagrajrao: The DEFT validator now delegates to `validate_configs(logging)` after checking the allocation contract, restoring DNS and rank-zero bind checks while preserving the normal TAO parameter-error path.

Revision: `9d3c8e46`.

nvidia_tao_pytorch/core/entrypoint.py
nnagrajrao: Updated the comment: on the DEFT path the spec, supported overrides, and environment must agree; no source silently takes precedence.

Revision: `9d3c8e46`.

vpraveen-nv (approved):
Verified all three round-2 items at 9d3c8e4: `_refinement_trainer_plugins()` + `plugins=` now live in this PR with `tests/distributed/test_deft_trainer_environment.py` covering the Slurm-env override; `_validate_native_dinov3_deft_launch` ends with `validate_configs(logging)` so the DNS/bind probes are back; precedence comment updated. Static checks/DCO green. Approving.

Cosmetic follow-up only: a DNS/port failure inside the DEFT path is reported as 'DINOv3 DEFT allocation override is invalid: Invalid master address ...' — the 'allocation override' wording is a bit misleading for a network check.

PR #133, opened 2026-09-10
[TAO-2529][Feature] Add manifest-backed DINOv3 refinement training

nnagrajrao:

Add DINOv3-owned file/tar/zip manifest loading, shard-local sampling, immutable-base candidate training, and deterministic checkpoint publication and retention.

**DEFT for SSL**

Within a repository, arrows are Git base dependencies and should be reviewed and merged left to right. Cross-repository links are runtime/API dependencies. This increment is based on `nnagrajrao/deft/01-foundations`, contains one signed-off commit, and is headed by `nnagrajrao/deft/02-manifest-training`.

Each refinement candidate must train from the original checkpoint over an exact cumulative manifest, resume safely within a round, and expose a verified teacher checkpoint.

### Related issues

DINOv3 training accepts an optional canonical Parquet manifest; directory-backed training remains supported and NVDINOv2 behavior is unchanged.

The final clean committed snapshots were archived and tested inside the approved TAO 7.2 rc36 containers:

- Final repository tips: Skill Bank `edef2e5d`, PyTorch `c22f3b80`, Data Services `7f2497e3`, Core `b3fd05f`.
- Source CI handoff: Skill Bank validation/hooks plus 27 tests; PyTorch hooks plus 323 passed and 4 expected weight-dependent skips; Data Services hooks plus 216 passed; Core hooks plus 19 passed.
- Broader QA: PyTorch 263 passed and 4 expected skips; Data Services 211 passed; Core 19 passed; Skill Bank 27 passed.
- Cross-repository dummy-data smoke: 5 passed, covering real image materialization/consumption, a DS LocalRunner process boundary into precomputed GRIT, exact embedding binding, schema equality, and DS/PyTorch attestation parity.
- Real Lightning smoke covered two local workers and two real nodes without torchrun.
- Independent adversarial panel re-review found no remaining source, contract, isolation, or test-coverage blocker.

Approved images were pinned to PyTorch `sha256:83eb5598411bf5a2b28f366e8861153e8da77f805edef0d200baa95bf3c5dc62` and DS `sha256:ab55b92d8c0eb570c29661dc8fe4a03dfe76cd5cc3a685ed77edcfc221c0f87a`. The source-free packaged probes fail only because rc36 predates the new PyTorch/DS modules and Core schema; rebuilding compatible release images remains a release-time requirement. No Dockerfile or dependency manifest changed.

Each GitLab integration branch is one signed-off commit over current `origin/main` (`0 1` ancestry), and its tree exactly matches the final GitHub stack tip. Production-scale throughput, real-data quality, full multi-node/FSDP training, and rebuilt packaged images remain outside this validation.

### Release note

```release-note
DINOv3 SSL supports canonical manifest-backed refinement training.
```

### Checklist

- [ ] Published release images contain the complete feature stack — pending normal TAO image publication.

### Notes for reviewers

Review base: `nnagrajrao/deft/01-foundations`; head: `nnagrajrao/deft/02-manifest-training`. This PR is ready for review and is not a draft. The scope is DINOv3 SSL DEFT only; other models keep their existing behavior.


nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+        raise ValueError("lr_scaling_rule must be sqrt_global_batch")
+    reference_world_size = int(lr_reference_world_size or world_size)
+    if reference_world_size <= 0:
+        raise ValueError("lr_reference_world_size must be positive")
+    lr_multiplier = math.sqrt(world_size / reference_world_size)
+    samples_per_rank = math.ceil(rows / world_size)
+    steps_per_pass = math.ceil(samples_per_rank / batch_size)
+    total_optimizer_steps = steps_per_pass * passes
+    if requested_checkpoint_unit == "epoch":
+        effective_checkpoint_interval = math.gcd(
```
vpraveen-nv: 🟠 **Major**: This still rewrites the user's `train.checkpoint_interval` to `gcd(interval, passes)` for the epoch unit. With `checkpoint_interval: 10, passes: 3` that becomes 1 → a full Lightning checkpoint plus student/teacher exports every epoch, i.e. *more* I/O, not 'preserved cadence' as the description claims. `DinoV3ModelCheckpoint(save_final_epoch=True)` already guarantees the terminal-epoch save off-cadence. Please drop the gcd and pass the requested interval through for both units, and add a test with a non-divisible interval (e.g. 10/3) asserting periodic saves plus the final epoch only.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+            requested_checkpoint_interval, passes
+        )
+    else:
+        effective_checkpoint_interval = min(
+            requested_checkpoint_interval, total_optimizer_steps
+        )
+    training["checkpoint_interval_unit"] = requested_checkpoint_unit
+    training["checkpoint_interval"] = effective_checkpoint_interval
+    warmup_steps = max(1, round(total_optimizer_steps * 0.10))
```
vpraveen-nv: 🟠 **Major**: `build_training_spec` silently overwrites `warm_up_steps` (10% of total) for lr/last-layer-lr/teacher-temperature, `max_decay_steps` for all five schedulers, and clamps `freeze_steps` to 10% — regardless of the base spec. Users are told this is 'a normal TAO DINOv3 spec', so they'll be surprised their schedule isn't honoured. Either make these explicit CLI knobs with the 10% behaviour as documented default, or only rescale when the base spec leaves them unset, and log every rewritten field. At minimum the docstring and README must list which fields are overridden.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+            expected_step=expected_step,
+        )
+        if recovered is None:
+            raise RuntimeError(
+                f"Final EMA teacher checkpoint is not loadable: {selected}"
+            ) from error
+        selected = recovered
+    temporary_checkpoint = output_dir / "checkpoint.tmp.pth"
+    temporary_checkpoint.unlink(missing_ok=True)
+    temporary_checkpoint.hardlink_to(selected)
```
vpraveen-nv: 🟠 **Major**: `hardlink_to` raises `OSError` (EPERM/ENOTSUP) on CIFS/SMB mounts — exactly where TAO experiment outputs live on our devboxes (`/media/scratch.metropolis*`) — and on any results dir on a different filesystem than the source. Please fall back to `shutil.copy2` to the tmp name + `replace` when `link()` fails, and record `checkpoint_publication: hardlink|copy` in the contract. The README caveat isn't a substitute — a user on scratch would just see a stack trace after a full training run.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint_retention.py
```diff
@@ -0,0 +1,41 @@
...
+
+from __future__ import annotations
+
+import os
+from pathlib import Path
+import re
+
+
+def _checkpoint_extent(path: Path) -> tuple[int, int, str]:
+    match = re.search(r"(?:epoch_)?(\d+)(?:_step_)?(\d+)\.pth$", path.name)
```
vpraveen-nv: 🟠 **Major**: This regex doesn't parse the real periodic filenames. For `model_003_00013.pth` (the `model_{epoch:03d}_{step:05d}` pattern the callback writes) `re.search` matches only the trailing step digits and returns `(1, 3)`; `model_010_00100.pth` → `(10, 0)`. It stays monotonic in step so pruning currently works by accident, but epoch ordering is wrong and any filename tweak breaks it. `test_periodic_checkpoint_retention_keeps_newest_families` masks this by fabricating `model_001_step_00001.pth`, which the callback never produces. Use an explicit per-family pattern (or reuse the correct regex in `train_cli._latest_resume_checkpoint`) and fix the test to use real names.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint_retention.py
```diff
@@ -0,0 +1,41 @@
...
+    if match is None:
+        return (-1, -1, path.name)
+    return (int(match.group(1)), int(match.group(2)), path.name)
+
+
+def prune_periodic_ssl_checkpoints(results_dir: str) -> None:
+    """Keep only the newest requested SSL checkpoint families when configured."""
+    raw_keep = os.environ.get("TAO_SSL_CHECKPOINT_KEEP_LAST_N")
```
vpraveen-nv: 🟡 **Minor**: Behaviour-changing knobs are being introduced as env vars (`TAO_SSL_CHECKPOINT_KEEP_LAST_N` here, `TAO_SHARD_SHUFFLE_WINDOW` and `TAO_ARCHIVE_HANDLE_CACHE` in `dataloader/dataset.py`). TAO convention is dataclass config fields (`train.checkpointer.*`, `dataset.*`) so they show up in `default_specs`, the API schema (tao-core #38) and `experiment.yaml`. Env vars are invisible to the microservice and to reproducibility. Please move these to the DINOv3 config or justify why they must be env-only.

nvidia_tao_pytorch/ssl/dinov3/dataloader/pl_dinov3_data_module.py
```diff
@@ -0,0 +1,143 @@
...
+from nvidia_tao_pytorch.ssl.nvdinov2.dataloader.transform import DinoV2Transform
+from nvidia_tao_pytorch.ssl.dinov3.dataloader.dataset import (
+    DinoV3Dataset,
+    ShardAwareDistributedSampler,
+)
+from nvidia_tao_pytorch.ssl.nvdinov2.dataloader.collate import DinoV2Collate
+from nvidia_tao_pytorch.core.distributed.comm import is_dist_avail_and_initialized
+
+
+class DinoV3DataModule(pl.LightningDataModule):
```
vpraveen-nv: 🟠 **Major**: This is a near-verbatim copy of `nvdinov2/dataloader/pl_dinov2_data_module.py` (~15 lines differ: manifest arg, sampler choice, `drop_last`, `persistent_workers`). Please subclass `DinoV2DataModule` and override `setup()`/`train_dataloader()` (or add a `_build_train_sampler()` hook) rather than forking — otherwise every future NVDINOv2 fix silently diverges here. Same for `DinoV3Dataset` vs `DinoV2Dataset`. 'NVDINOv2 files byte-for-byte unchanged' is achievable via inheritance; it doesn't require duplication.

nvidia_tao_pytorch/ssl/dinov3/dataloader/dataset.py
```diff
@@ -0,0 +1,382 @@
...
+        else:
+            padding = epoch_size - len(indices)
+            indices += (indices * ((padding // len(indices)) + 1))[:padding]
+        if len(indices) != epoch_size:
+            raise RuntimeError("Shard-aware sampler produced an invalid partition")
+        return iter(indices)
+
+
+class DinoV3Dataset(Dataset):
+    """Dataset for NVDINOv2 to manage and transform image data for training, with support for various image formats."""
```
vpraveen-nv: 🟡 **Minor**: Copy-paste residue: docstring says 'Dataset for NVDINOv2', header year is 2024 on a new file, and `_list_images`/`__getitem__`/predict-path logic duplicate `DinoV2Dataset`. Suggest `class DinoV3Dataset(DinoV2Dataset)` that only adds `manifest_path`, `_read_manifest`, `_resolved_path`, `_archive` and `sampling_group`. Also `_read_manifest` requires only `storage_type/path` while `train_cli.build_training_spec` demands `sample_id` too — pick one canonical column set.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+    contract["commit_record"] = str(commit_path)
+    _atomic_write_text(
+        contract_path, json.dumps(contract, indent=2, sort_keys=True) + "\n"
+    )
+    commit_sha256 = _validate_training_commit(output_dir, contract)
+    _publish_or_verify_success(output_dir, commit_sha256)
+    return checkpoint
+
+
+def main(argv: list[str] | None = None) -> int:
```
vpraveen-nv: 🟠 **Major**: Design: this is a second, argparse-based entrypoint that spawns the real TAO entrypoint via `subprocess.run`, wraps it in a ~30-field provenance contract (including SHA-256 of `sys.executable` and nine source files), a prepared-marker handshake, a 300 s polling loop, and three recovery paths. That's a lot of machinery in tao-pytorch that doesn't follow `entrypoint/ -> scripts/ -> config/` and can't be invoked as `dinov3 <subtask>`. Since the DS controller (#40) is the only caller, `build_training_spec` belongs in the controller or a small `dinov3 train` config option, and the contract should shrink to inputs that actually affect reproducibility (spec, manifest, parent ckpt, resume ckpt, final ckpt). Hashing the Python binary on every launch isn't meaningful provenance and costs real I/O on shared mounts.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+            r"model_(?:epoch_)?(\d+)_(?:step_)?(\d+)\.pth", path.name
+        )
+        if match is not None:
+            return (1, int(match.group(1)), int(match.group(2)), path.name)
+        return (0, 0, 0, path.name)
+
+    candidates = sorted(candidates, key=position, reverse=True)
+    for candidate in candidates:
+        try:
+            payload = _load_torch_checkpoint(candidate, weights_only=False)
```
vpraveen-nv: 🟡 **Minor**: `_latest_resume_checkpoint` does `torch.load(..., weights_only=False)` on every `model_*.pth` and `dinov3_model_latest.pth` in the results dir just to check keys. Anything dropped into `output_dir` matching `model_*.pth` gets unpickled with full code execution. Try `weights_only=True` first and only fall back for the Lightning payload, and restrict candidates to the `dinov3_model_latest.pth` target plus the newest `model_*` by name.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+    payload = _load_torch_checkpoint(full_state, weights_only=False)
+    if (
+        int(payload.get("epoch", -1)) != expected_epoch or
+        int(payload.get("global_step", -1)) != expected_step
+    ):
+        return None
+    state_dict = payload.get("state_dict")
+    if not isinstance(state_dict, dict):
+        return None
+    teacher = {
```
vpraveen-nv: 🟡 **Minor**: The 'strip `teacher.backbone.`, drop `dino_head.*` and `mask_token`' logic duplicates `DinoV3ModelCheckpoint._save_checkpoint` (checkpoint.py ~L100-117). Please factor one helper (e.g. `export_backbone_state(state_dict, prefix)`) in `model/checkpoint.py` and call it from both places so the recovered teacher can never differ from the callback-exported one.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+    parquet = pq.ParquetFile(manifest_path)
+    missing = {"sample_id", "storage_type", "path"}.difference(
+        parquet.schema_arrow.names
+    )
+    if missing:
+        raise ValueError(
+            f"Training manifest is missing canonical columns: {sorted(missing)}"
+        )
+    resume_checkpoint = _latest_resume_checkpoint(Path(output_dir).resolve())
+    training["resume_training_checkpoint_path"] = (
```
vpraveen-nv: 🟡 **Minor**: When no *loadable* full-state checkpoint is found this writes `resume_training_checkpoint_path: null`, but `initialize_train_experiment` then falls back to `get_latest_checkpoint(results_dir)`, which will happily pick up the `dinov3_model_latest.pth` symlink you just rejected (e.g. truncated) and crash inside `trainer.fit`. Either pass an explicit sentinel that disables the TAO fallback, or remove/rename the rejected checkpoint before launch.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+        None if resume_checkpoint is None else str(resume_checkpoint)
+    )
+    rows = int(parquet.metadata.num_rows)
+    batch_size = int(dataset["batch_size"])
```
vpraveen-nv: 🟡 **Minor**: `dataset["batch_size"]` raises a bare `KeyError` if the base spec relies on the dataclass default (every other field here uses `.get(..., default)`). Read the default from `DINOv3DatasetConfig` or raise a `ValueError` naming the field.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
```diff
@@ -0,0 +1,1106 @@
...
+    parser.add_argument("--num-nodes", type=int)
+    parser.add_argument("--gpus-per-node", type=int)
+    parser.add_argument(
+        "--lr-scaling-rule",
+        choices=("sqrt_global_batch",),
+        default="sqrt_global_batch",
+    )
+    parser.add_argument("--lr-reference-world-size", type=int)
+    parser.add_argument(
+        "--checkpoint-policy",
```
vpraveen-nv: ❔ **Question**: `--checkpoint-policy` has three choices but no code path reads it — it's only echoed into the contract/digest. Is `previous_round_checkpoint` supposed to change what `--checkpoint` means? If it's purely a controller-owned label, drop it from this CLI so the pytorch side doesn't carry a no-op flag.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, *args, save_final_epoch=False, **kwargs):
+        """Optionally publish the final manifest-training extent off cadence."""
+        super().__init__(*args, **kwargs)
+        self.save_final_epoch = save_final_epoch
+
+    def on_train_epoch_end(self, trainer, pl_module):
+        """Complete manifest training without increasing periodic save frequency."""
+        super().on_train_epoch_end(trainer, pl_module)
+        if (
```
vpraveen-nv, line 68: 🟡 **Minor**: The `pl_model.configure_callbacks` docstring says these callbacks 'do not depend on Lightning internals', but this override calls four private `ModelCheckpoint` members (`_last_global_step_saved`, `_should_skip_saving_checkpoint`, `_monitor_candidates`, `_save_topk_checkpoint`/`_save_last_checkpoint`). Acceptable given we pin `pytorch-lightning==2.6.1`, but fix the docstring and add a comment noting the pin so the next Lightning bump re-checks this.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
```diff
@@ -0,0 +1,136 @@
...
+
+        Args:
+            trainer (pl.Trainer): The PyTorch Lightning trainer instance, providing access to model and training information.
+            filepath (str): The file path where the checkpoint will be saved.
+        """
+        # Call the original save_checkpoint method to save the checkpoint as usual
+        trainer.save_checkpoint(filepath, self.save_weights_only)
+        # FSDP state-dict extraction is collective, so every rank must enter it.
+        # Only global rank zero may publish the converted checkpoint files.
+        state_dict = trainer.lightning_module.state_dict()
```
vpraveen-nv: ❔ **Question**: Under `FSDPStrategy(state_dict_type="full")`, `trainer.save_checkpoint` gathers via `strategy.lightning_module_state_dict()` (rank0-only, CPU-offloaded), but this direct `trainer.lightning_module.state_dict()` uses FSDP's default FULL_STATE_DICT with no offload and *all* ranks materialising the full model on GPU — for ViT-L/H+ that's the OOM FSDP is meant to avoid. Since DINOv3 now owns this callback, consider `trainer.strategy.lightning_module_state_dict()` (empty on non-zero ranks, matching the early return below).

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
```diff
@@ -952,46 +957,33 @@ def configure_callbacks(self):
...
-                ))
-                logging.info(
-                    "DINOv3: checkpointing every %d steps (checkpoint_interval_unit='step').",
-                    interval,
-                )
-            else:
-                rebuilt.append(callback)
-        return rebuilt
+        # Construct DINOv3-owned callbacks without changing NVDINOv2 class attributes.
+        periodic = DinoV3ModelCheckpoint(
```
vpraveen-nv, line 1012: 🟡 **Minor**: `save_final_epoch` is gated on `dataset.train_manifest` being set. A directory-backed DINOv3 run with `num_epochs` not divisible by `checkpoint_interval` has the same 'no terminal checkpoint' problem. Unless there's a reason to keep legacy behaviour for directory datasets, enable it unconditionally (it's a no-op when the last epoch is on cadence).

nvidia_tao_pytorch/ssl/dinov3/dataloader/dataset.py
```diff
@@ -0,0 +1,382 @@
...
+        key = (storage_type, str(path))
+        archive = self._archive_handles.pop(key, None)
+        if archive is None:
+            if storage_type == "tar":
+                if path.suffix.lower() != ".tar":
+                    raise ValueError(
+                        "Random archive-member training requires uncompressed .tar "
+                        f"shards; stage or reshard {path}"
+                    )
+                archive = tarfile.open(path, mode="r:")
```
vpraveen-nv: ⚪ **Nit**: `tarfile.open(...).getmembers()` walks the entire shard on every (re)open, per worker, and again after every LRU eviction (`TAO_ARCHIVE_HANDLE_CACHE=8`). For multi-GB shards with >8 shards per rank this becomes the dominant I/O. Consider building the member index once in the main process (or requiring a sidecar index) and pickling it into workers.

tests/ssl_unit_test/dinov3/test_refinement_manifest.py
```diff
@@ -0,0 +1,1492 @@
...
+
+
+def _transform(_image):
+    return {"global_crops": ["global"], "local_crops": ["local"]}
+
+
+@pytest.mark.parametrize(("relative_path", "reader", "count"), [
+    ("dinov3/dataloader/dataset.py", "read_table", 1),
+])
+def test_deft_parquet_reads_disable_only_background_prefetch(
```
vpraveen-nv: ⚪ **Nit**: Asserting on the *source text* (AST) of `dataset.py` to pin `pre_buffer=False` is brittle and doesn't test behaviour. If you want a guard, monkeypatch `pq.read_table` and assert the kwargs at runtime.

vpraveen-nv (commented):
initial comments.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: Fixed by deleting train_cli and preserving the configured interval in the DINOv3 Lightning callback. Real Lightning regressions cover 10/3 epoch cadence plus 10/2/3 step cadences and verify a single off-cadence terminal save.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: Fixed by moving spec composition to the DS native action and preserving the resolved base-spec scheduler fields. Only round-owned inputs (manifest, passes, resources, original checkpoint, and auto-resume=false) are set; the contract records scheduler_policy=preserve_resolved_base_spec.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: The hardlink publication path was removed with train_cli. Final checkpoints are published by the DINOv3 Lightning callback with atomic temporary-file replacement, so CIFS/cross-filesystem hardlink behavior is no longer involved.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint_retention.py
nnagrajrao: Fixed. Retention now full-matches the actual model_epoch_NNN_step_NNNNN and student/teacher family names, ignores unrelated/versioned files, and the tests use real callback filenames.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint_retention.py
nnagrajrao: Fixed. archive_cache_size and shard_shuffle_window are dataset dataclass fields; checkpoint_keep_last_n and auto_resume are train dataclass fields. They are present in the PyTorch/Core schemas and resolved experiment config.

nvidia_tao_pytorch/ssl/dinov3/dataloader/pl_dinov3_data_module.py
nnagrajrao: Fixed through inheritance. DinoV3DataModule subclasses DinoV2DataModule and overrides only manifest setup plus loader differences; DinoV3Dataset subclasses DinoV2Dataset and delegates storage reads to the shared DINOv3 locator reader. NVDINOv2 sources remain unchanged.

nvidia_tao_pytorch/ssl/dinov3/dataloader/dataset.py
nnagrajrao: Fixed. The file is DINOv3-owned, subclasses DinoV2Dataset, has the corrected header/docstrings, and enforces one canonical sample_id/storage_type/path/member contract.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: Agreed and redesigned. train_cli and its subprocess/polling/provenance machinery were deleted. DS writes a small structured spec and launches the standard dinov3 train action; Lightning owns workers and the sealed contract contains only reproducibility-relevant inputs.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: Removed with train_cli. Resume discovery now uses TAO's normal Lightning checkpoint path and DEFT explicitly sets auto_resume=false for a fresh candidate.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: Fixed by having the DINOv3 checkpoint callback own one _backbone_state_dict helper and all final teacher publication; the deleted wrapper no longer has a second stripping implementation.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: Fixed. DEFT sets train.auto_resume=false in the generated native spec, so initialize_train_experiment cannot fall back to an unrelated results-dir checkpoint.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: Fixed through OmegaConf structured defaults. DS reads spec.dataset.batch_size after composing the user spec over ExperimentConfig, so an omitted value receives the declared default rather than raising KeyError.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
nnagrajrao: checkpoint_policy is now controller-owned metadata and is not a PyTorch CLI option. The native training spec receives the already selected parent checkpoint, while the DS contract records the policy.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
nnagrajrao: Updated. The callback no longer claims independence from Lightning internals; its terminal-save behavior is isolated to DINOv3 and covered with real Lightning epoch/step cadence tests.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
nnagrajrao: Fixed. Export uses trainer.strategy.lightning_module_state_dict() collectively, and only global rank zero performs atomic publication.

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
nnagrajrao: Fixed. save_final_epoch is enabled for the DINOv3 periodic callback unconditionally; it is a no-op when the final extent was already saved on cadence.

nvidia_tao_pytorch/ssl/dinov3/dataloader/dataset.py
nnagrajrao: The shared LocalImageReader now owns fork/spawn-safe handles and indexes, clears inherited state, and evicts the paired index with the handle. The cache remains deliberately bounded; a persistent sidecar index would be a separate format change.

tests/ssl_unit_test/dinov3/test_refinement_manifest.py
nnagrajrao: Fixed. Runtime spies now assert pre_buffer=False on the executed Parquet read paths; the source-text assertion was removed.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Verified — `train_cli.py` is deleted and `configure_callbacks` passes the configured interval straight through; `test_real_lightning_callback_saves_terminal_extent_once` (10/3 epochs) and the step-cadence test confirm a single off-cadence terminal save. Resolved.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Verified against DS #40's spec composer: only `results_dir`, `train_manifest`, `images_dir` fallback, `num_epochs`, node/GPU fields, `pretrained_model_path`, `resume_training_checkpoint_path=None`, `auto_resume=False` are set; scheduler fields come from the structured merge of the user spec. Resolved.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Moot — no `os.link`/`hardlink_to` anywhere in the DINOv3 tree; publication is `_atomic_torch_save` (temp + `replace`) on global rank 0. Resolved (small nit on the temp name left as a new comment).

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint_retention.py
vpraveen-nv: Verified — `_CHECKPOINT_PATTERNS` `fullmatch` `model_epoch_NNN_step_NNNNN.pth` and the student/teacher/student_ema families; non-matching names raise instead of mis-parsing, and the retention test uses real callback filenames. Resolved.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint_retention.py
vpraveen-nv: Verified — `dataset.archive_cache_size`, `dataset.shard_shuffle_window`, `train.checkpoint_keep_last_n`, `train.auto_resume` are dataclass fields here and byte-identical in tao-core #38; no `TAO_*` env knobs remain in `ssl/dinov3`. Resolved.

nvidia_tao_pytorch/ssl/dinov3/dataloader/pl_dinov3_data_module.py
vpraveen-nv: Verified — `DinoV3DataModule(DinoV2DataModule)` overrides `setup`/`train_dataloader`/`predict_dataloader` only; `nvdinov2/` is untouched in the diff. Resolved.

nvidia_tao_pytorch/ssl/dinov3/dataloader/dataset.py
vpraveen-nv: Verified — 2026 header, DINOv3 docstrings, `DinoV3Dataset(DinoV2Dataset)` with only the manifest additions, and the `sample_id/storage_type/path(+member)` column set matches DS #40's producer check. Resolved.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Verified — the wrapper, subprocess/polling loop and provenance hashing are gone; DS writes `refinement_input.yaml` and launches the standard `dinov3 train` under `TAO_REFINEMENT_LIGHTNING_LAUNCH=1`. Thanks for taking the bigger redesign. Resolved.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Moot — no `weights_only=False` directory scan remains; resume goes through `initialize_train_experiment(auto_resume=cfg.train.auto_resume)` and DEFT sets it false. Resolved.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Verified — single `_backbone_state_dict` helper in `model/checkpoint.py`, used for both student and teacher exports. Resolved.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Verified — `initialize_train_experiment` only calls `get_latest_checkpoint` when `auto_resume` is true, DS sets it false, and `test_dino_opt_ins_allow_terminal_save_without_directory_resume` asserts the scan is never invoked. Resolved.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Verified in DS #40 `resolved_training_batch_size` — reads `spec.dataset.batch_size` off `OmegaConf.merge(OmegaConf.structured(ExperimentConfig()), user_spec)`, so the dataclass default applies. Resolved.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/train_cli.py
vpraveen-nv: Verified — no `checkpoint_policy` on the PyTorch side; DS owns it. Resolved.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
vpraveen-nv: Docstring is fixed, thanks. The callback still calls `_last_global_step_saved`, `_should_skip_saving_checkpoint`, `_monitor_candidates`, `_save_topk_checkpoint`, `_save_last_checkpoint`; a one-line comment in `DinoV3ModelCheckpoint.on_train_epoch_end` noting these are private `ModelCheckpoint` members validated against the pinned Lightning version would close this. Not blocking.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
vpraveen-nv: Verified — `checkpoint.py:65` uses `trainer.strategy.lightning_module_state_dict()` collectively and returns before export on non-zero ranks; `test_converted_checkpoints_are_published_only_by_global_rank_zero` covers it. Resolved (not runtime-tested under a real `FSDPStrategy`, but the accessor change is unambiguous).

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
vpraveen-nv: Verified — `save_final_epoch=True` unconditionally in `configure_callbacks`, and `allow_off_cadence_final_checkpoint=True` relaxes the core assert only for DINOv3. Resolved.

nvidia_tao_pytorch/ssl/dinov3/dataloader/dataset.py
vpraveen-nv: Acknowledged — the bounded `LocalImageReader` cache with paired index eviction is fine for now; a sidecar index is a format change for later. Resolved.

tests/ssl_unit_test/dinov3/test_refinement_manifest.py
vpraveen-nv: The AST/source-text assertion is gone, which is what I asked for. Resolved.

vpraveen-nv (commented):
Thanks for the substantial rework — I re-read the full PR at 86eedc5. `train_cli.py` is gone; spec composition moved to DS #40, `DinoV3DataModule`/`DinoV3Dataset` subclass the NVDINOv2 classes, retention `fullmatch`es real callback names, the new knobs are dataclass fields identical to tao-core #38, the callback exports via `trainer.strategy.lightning_module_state_dict()` with rank-0 atomic publication, and `save_final_epoch` is unconditional with real Lightning cadence tests. All 18 earlier threads are resolved or moot (one partial: the Lightning-internals note) — replied inline.

Fresh pass: one **major** — the manifest loader switched to `BatchSampler(drop_last=False)`, which yields a ragged final batch on every rank each epoch; DINOv2's KoLeo loss is degenerate at B=1 and centering statistics suffer, so I'd pad the sampler's `epoch_size` to a batch multiple and keep `drop_last=True`. Minors: manifest column docs (missing `sample_id`/`replay_repeat`, no README section), empty `images_dir` resolving relative rows against CWD, fixed temp name in `_atomic_torch_save`, GRIT-only helpers landing here without a caller, and no tao-core parity test that actually runs. Requesting changes for the `drop_last` point; everything else is approve-with-comments.

nvidia_tao_pytorch/ssl/dinov3/dataloader/pl_dinov3_data_module.py
```diff
@@ -0,0 +1,93 @@
...
+                    ),
+                    num_workers=0,
+                    collate_fn=DinoV2Collate(patch_size=self.patch_size),
+                    pin_memory=True,
+                    persistent_workers=False,
+                )
+            return super().train_dataloader()
+        return DataLoader(
+            self.train_dataset,
+            batch_sampler=BatchSampler(self.train_sampler, self.batch_size, drop_last=False),
```
vpraveen-nv: 🟠 **Major**: `drop_last=False` here changes SSL training dynamics, not just I/O. `ShardAwareDistributedSampler` gives every rank exactly `num_samples = ceil(N / world)` indices, so each epoch every rank ends with the same ragged batch of `num_samples % batch_size` samples. DINOv2's loss assumes full batches: `KoLeoLoss` runs per global-crop chunk and with a local B=1 the nearest neighbour is the sample itself → `-log(2e-8) ≈ 17.7` per crop on the last step of every epoch; DINO/iBOT centering and Sinkhorn statistics are also computed over a tiny batch. NVDINOv2 keeps `drop_last=True` for exactly this reason. Suggestion that still honours 'every manifest row is seen': have the sampler pad `epoch_size` up to a multiple of `batch_size` (it already pads to `num_samples`, so this is a one-line change to the target size plus passing `batch_size` in) and go back to `drop_last=True`. DS #40's `steps_per_pass = ceil(ceil(rows/world)/batch_size)` stays exact under that scheme.

nvidia_tao_pytorch/config/dinov3/default_config.py
```diff
@@ -466,6 +466,36 @@ class DINOv3TransformConfig(NVDINOv2TransformConfig):
 class DINOv3DatasetConfig(NVDINOv2DatasetConfig):
     """DINOv3 dataset config (reuses nvdinov2 dataset, v3 transform defaults)."""
 
+    train_manifest: Optional[str] = STR_FIELD(
+        value=None,
+        default_value=None,
+        default_type=None,
+        description=(
+            "Optional Parquet training manifest. Each row must contain an absolute or "
```
vpraveen-nv: 🟡 **Minor**: This description is the only user-facing documentation of the manifest contract in tao-pytorch and it doesn't match what `_read_manifest` enforces: `sample_id` is required (missing/null/empty/duplicate all raise) but isn't mentioned, `member` is required for `tar`/`zip` rows, and the optional `replay_repeat` column isn't documented anywhere in this repo. Please list the columns here and add a short 'Manifest-backed training' section to `ssl/dinov3/README.md` covering `train_manifest`, `auto_resume`, `checkpoint_keep_last_n`, `archive_cache_size`, `shard_shuffle_window` (the checklist ticks documentation but no README/spec file is in the diff). Keep the string identical to tao-core #38's copy.

nvidia_tao_pytorch/ssl/dinov3/dataloader/dataset.py
```diff
@@ -0,0 +1,235 @@
...
+        self.manifest_path = Path(manifest_path) if manifest_path else None
+        if self.manifest_path is None:
+            super().__init__(root=root, transform=transform, train=train, **kwargs)
+            self._reader = None
+            return
+        if transform is None:
+            raise ValueError("Transform must be specified")
+        if not self.manifest_path.is_file():
+            raise FileNotFoundError(f"train_manifest does not exist: {self.manifest_path}")
+        self.root = Path(root)
```
vpraveen-nv: 🟡 **Minor**: In manifest mode `root` comes from `dataset.train_dataset.images_dir`, whose dataclass default is `""`, so `Path("")` is `.` and any relative `path` row silently resolves against the process CWD. DS #40 works around it by setting `images_dir="/"` when empty, which is a hint the contract is under-specified. Suggest failing fast: if `root` is empty and any row's resolved path is relative, raise a `ValueError` naming `dataset.train_dataset.images_dir` (or resolve relative rows against the manifest's own directory and say so in the description).

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
```diff
@@ -0,0 +1,114 @@
...
+
+from nvidia_tao_pytorch.ssl.nvdinov2.model.pl_model import CustomModelCheckpoint
+from nvidia_tao_pytorch.ssl.dinov3.model.checkpoint_retention import prune_periodic_ssl_checkpoints
+from nvidia_tao_pytorch.core.callbacks.model_checkpoint import TAOExceptionCheckpoint
+
+
+def _atomic_torch_save(value, path: str) -> None:
+    """Publish a checkpoint export without exposing a partial final file."""
+    destination = Path(path)
+    temporary = destination.with_name(f".{destination.name}.tmp")
```
vpraveen-nv: 🟡 **Minor**: Same fixed-temp-name pattern I flagged on the spec writer: two writers in one `results_dir` (retry racing a stale process, or two runs pointed at the same directory) can `unlink` each other's `.student_epoch_...pth.tmp` before `replace()` and one dies with `FileNotFoundError` after a full epoch. You already have the right recipe in `utils/runtime_spec.py:24-40` (`tempfile.mkstemp(dir=..., prefix=...)`, fsync, `os.replace`, unlink only your own path) — please reuse it here so the two atomic publishers don't diverge.

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
```diff
@@ -795,6 +840,38 @@ def restore_pretrained_weights(self, preloaded_state_dict=None):
         if getattr(self, 'gram_teacher', None) is not None:
             self._sync_gram_teacher()
 
+    @classmethod
+    def load_backbone_weights(cls, backbone, checkpoint_path):
```
vpraveen-nv: 🟡 **Minor**: `load_backbone_weights` has no caller in this PR (nor does the standalone `build_backbone` classmethod beyond the `_make_backbone` refactor) — the first production user is #134's `grit_pipeline._build_backbone`. `tests/ssl_unit_test/dinov3/test_attention_gate.py` is likewise GRIT-motivated. Same increment-hygiene point as on #132: either move these to #134 with their caller, or add a sentence to this PR's description so reviewers know why a 'manifest training' PR ships an inference-side loader.

nvidia_tao_pytorch/config/dinov3/default_config.py
```diff
@@ -539,6 +569,27 @@ class DINOv3TrainExpConfig(NVDINOv2TrainExpConfig):
         display_name="logging interval",
         popular="no",
     )
+    checkpoint_keep_last_n: int = INT_FIELD(
```
vpraveen-nv, line 574: 🟡 **Minor**: Cross-repo schema parity for these five new fields is enforced nowhere that runs by default: tao-core #38's `test_refinement_config.py` `importorskip`s `nvidia_tao_pytorch` (so it skips in tao-core CI), and nothing here imports `nvidia_tao_core`. The tao-pytorch container *does* ship tao-core, so a small test in `tests/ssl_unit_test/dinov3/` that `pytest.importorskip("nvidia_tao_core.config.dinov3.default_config")` and asserts the `DINOv3DatasetConfig`/`DINOv3TrainExpConfig` field names + defaults match would actually execute in our pipeline and catch drift.

nvidia_tao_pytorch/ssl/dinov3/scripts/train.py
```diff
@@ -46,13 +49,42 @@ def _resolve_strategy(experiment_config):
...
+
+    The node-level adapter deliberately launches one ordinary Python process per
+    node. Lightning creates the local workers later, so its rank-zero helper has
+    not yet learned the global rank when ``monitor_status`` emits STARTED. Seed
+    that helper from the adapter's node coordinates without setting ``RANK`` (an
+    external-worker launch signal to Lightning).
+    """
+    if os.environ.get("TAO_REFINEMENT_LIGHTNING_LAUNCH") != "1":
+        return
+    node_rank = int(os.environ.get("NODE_RANK", "0"))
```
vpraveen-nv: ⚪ **Nit**: `int(os.environ.get("NODE_RANK", "0"))` is the same empty-string `ValueError` pattern you fixed in `runtime_spec._nonzero_rank`. Safe today only because `launch()` rejects empty `NODE_RANK`/`NUM_GPU_PER_NODE`, but this module can be run directly; `int(os.environ.get(name) or "0")` for the three reads keeps it consistent.

nvidia_tao_pytorch/ssl/dinov3/dataloader/pl_dinov3_data_module.py
nnagrajrao: The manifest sampler now pads each rank to a complete batch, and the loader uses `drop_last=True`. This preserves full-batch SSL loss behavior while visiting every manifest row; the manifest regressions cover padded multi-rank coverage.

Revision: `a4f8d829`.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
nnagrajrao: Checkpoint exports and runtime-spec publication now share the DINOv3 atomic publisher: unique same-directory temporary files, file fsync, replace, directory fsync, and cleanup of only the writer's own temporary file. Concurrency/failure regression tests pass.

Revision: `a4f8d829`.

nvidia_tao_pytorch/config/dinov3/default_config.py
nnagrajrao: Expanded the identical PyTorch/Core manifest descriptions to include sample_id, file/tar/zip locators, archive member, and replay_repeat. Added the manifest-training README section covering resume, retention, archive cache, and shard shuffle.

Revision: `a4f8d829`.

nvidia_tao_pytorch/ssl/dinov3/dataloader/dataset.py
nnagrajrao: Relative row paths now resolve against the manifest directory when images_dir is empty. The contract is documented in both schema copies, and manifest-loading tests cover the behavior.

Revision: `a4f8d829`.

nvidia_tao_pytorch/ssl/dinov3/model/pl_model.py
nnagrajrao: Moved `load_backbone_weights` into #134 alongside its first inference caller. The backbone factory remains in #133 because `_make_backbone` uses it for training; its attention-configuration test covers that shared construction path.

Revision: `a4f8d829`.

nvidia_tao_pytorch/config/dinov3/default_config.py
nnagrajrao: Added parity tests to the PyTorch suite and made them mandatory in the cross-repo container harness via `TAO_DEFT_REQUIRE_SCHEMA_PARITY=1`. Standalone tests explicitly skip incompatible old companion packages instead of falsely passing; the combined run checks fields, defaults, and metadata with the candidate Core source.

Revision: `a4f8d829`.

nvidia_tao_pytorch/ssl/dinov3/scripts/train.py
nnagrajrao: All three direct train rank reads now use `os.environ.get(name) or "0"`, consistent with runtime-spec publication.

Revision: `a4f8d829`.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
nnagrajrao: Added the explicit private-ModelCheckpoint API note and pinned Lightning version in on_train_epoch_end; real callback cadence tests must be rerun when the dependency changes.

Revision: `a4f8d829`.

nvidia_tao_pytorch/ssl/dinov3/model/checkpoint.py
```diff
@@ -0,0 +1,145 @@
...
+        self._teacher_export = None
+        if self.publish_terminal_manifest and trainer.is_global_zero:
+            (Path(trainer.default_root_dir) / "terminal_teacher.json").unlink(missing_ok=True)
+
+    def on_train_end(self, trainer, pl_module):
+        """Publish the authoritative teacher extent after successful training."""
+        super().on_train_end(trainer, pl_module)
+        if not self.publish_terminal_manifest or not trainer.is_global_zero:
+            return
+        if not self._teacher_export or self._teacher_export["global_step"] != trainer.global_step:
```
vpraveen-nv, line 52: 🟡 **Minor**: New in this revision: `on_train_end` raises `ValueError("DINOv3 training ended without a terminal teacher export")` whenever `publish_terminal_manifest` is set and `_teacher_export` is None. `_save_checkpoint` only records `_teacher_export` for `name == "teacher"`, but in distill mode `names = ("student", "student_ema")`, so `dataset.train_manifest` + `model.distill.enable=true` (pl_model sets `publish_terminal_manifest=bool(train_manifest)` unconditionally) trains to completion and then fails here. Either record `student_ema` as the teacher in distill mode, gate `publish_terminal_manifest` on `not distill.enable`, or reject the combination up front. Follow-up, not blocking — DEFT doesn't set distill.

vpraveen-nv (approved):
Verified all seven round-2 items at a4f8d82, including the major: `ShardAwareDistributedSampler` now takes `batch_size`, pads `num_samples` to a batch multiple, the loader uses `drop_last=True`, and `test_manifest_sampler_pads_full_batches_without_losing_rows` covers (5,1,2)/(11,3,3)/(2,4,2). Manifest columns documented in both schema copies + README section; relative locators resolve against the manifest dir with a chdir test; shared `utils/atomic.py` with a two-writer test; `load_backbone_weights` moved to #134; `test_schema_parity.py` actually imports tao-core; env reads use `or "0"`. Static checks/DCO green. Approving.

Non-blocking follow-ups (new `terminal_teacher.json` scope introduced in this revision):
- `train_manifest` + `model.distill.enable=true` trains to completion and then fails in `on_train_end` (see inline) — DEFT never sets distill, so not a blocker.
- Number the new 'Manifest-backed training' README section once #134 lands.
- Wire `TAO_DEFT_REQUIRE_SCHEMA_PARITY=1` into the tao-pytorch container test job once tao-core #38 is in the shipped image so parity is mandatory rather than skip-on-old-core.

PR #134, opened 2026-09-10
[TAO-2529][Feature] Add DINOv3 GRIT scoring and native TAO action

nnagrajrao:

Add deterministic DINOv3 GRIT scoring and native scoring/training actions, with fail-fast cohort, device, scratch, and implementation-attestation validation while retaining pure Lightning orchestration.

**DEFT for SSL**

Within a repository, arrows are Git base dependencies and should be reviewed and merged left to right. Cross-repository links are runtime/API dependencies. This increment is based on `nnagrajrao/deft/02-manifest-training`, contains one signed-off commit, and is headed by `nnagrajrao/deft/03-grit-scoring`.

The DS-owned controller needs a sealed DINOv3-native scoring leaf that produces weakness rankings and embeddings without introducing torchrun or a second orchestration layer.

### Related issues

DINOv3 exposes native GRIT scoring and refinement action entrypoints; no other model is opted in.

Review revision: `cca70c6022c4b253856a1f29615d8236d1444dd5`. Clean committed snapshots were archived into the approved TAO 7.2 rc36 containers; repository pre-commit hooks ran without SKIP exemptions (including secret scanning where configured).

Full PyTorch stack: 335 tests passed with four missing-external-weight skips. Includes a real tiny-ViT CPU `extract_grit_scores` end-to-end test and mapped-status/attestation checks; this is not a full training-quality experiment.

Full-stack local source checks: Skill Bank 30 passed; PyTorch 335 passed / 4 missing-weight skips; Data Services 235 passed / 1 absent-cuVS skip; Core 20 passed. These are the selected feature/shared-surface tests and repository hooks, not a claim that every Jenkins job or every TAO model was tested.

Image IDs: PyTorch `sha256:83eb5598411bf5a2b28f366e8861153e8da77f805edef0d200baa95bf3c5dc62`; DS `sha256:ab55b92d8c0eb570c29661dc8fe4a03dfe76cd5cc3a685ed77edcfc221c0f87a`.

Release gates remain open: source-free packaged probes fail because rc36 lacks the new feature modules/schema. Independently, the unchanged DS image fails BF16 convolution forward/backward preflight on this RTX A6000; the same operation passes in the unchanged PyTorch image. Rebuilding alone is not proven to fix this. No Dockerfile or dependency manifest changed. Full DEFT training, production throughput/quality, rebuilt-image validation, and approved cuVS packaging are not certified by these tests.

The GitLab integration commit is one signed-off commit over the fetched actual GitHub main (`0 1`), with a tree matching the GitHub stack tip. Each PR remains one signed-off commit above its stated base.

### Release note

```release-note
DINOv3 SSL adds native GRIT scoring for DEFT refinement.
```

### Checklist

- [ ] Published release images contain the complete feature stack — blocked on compatible image publication and DS compute preflight.

### Notes for reviewers

Review base: `nnagrajrao/deft/02-manifest-training`; head: `nnagrajrao/deft/03-grit-scoring`. This source increment is available for review; it is not release-runtime approval. The scope is DINOv3 SSL DEFT only; other models keep their existing behavior.
### Explicit follow-up (TAO-2529)

- [ ] Reduce GRIT reference-pool work: skip settling for reference-only rows and reuse first-pass canonical CLS instead of re-forwarding queries. Correctness is unchanged; no throughput improvement is claimed for this deferred optimization.


nvidia_tao_pytorch/ssl/dinov3/scripts/grit_score.py
```diff
@@ -0,0 +1,50 @@
...
+from nvidia_tao_pytorch.core.hydra.hydra_runner import hydra_runner
+from nvidia_tao_pytorch.core.tlt_logging import obfuscate_logs
+from nvidia_tao_pytorch.ssl.dinov3.data_refinement.cli import run
+
+
+spec_root = Path(__file__).resolve().parents[1] / "experiment_specs"
+
+
+@hydra_runner(config_path=str(spec_root), config_name="experiment_spec", schema=ExperimentConfig)
+def main(cfg: ExperimentConfig) -> None:
```
vpraveen-nv, line 22: 🟠 **Major**: Adding a module under `scripts/` registers `grit_score` as a `dinov3` subtask (via `get_subtasks`), so the root `README.md` command table must be regenerated: `python tools/update_readme_supported_commands.py --check` fails at this head (expected row: `convert, default_specs, export, grit_score, inference, train`). Please run the tool and commit the README change.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
```diff
@@ -0,0 +1,194 @@
...
+        )
+    commit_path = output_dir / "score_commit.json"
+    _atomic_json(output_dir / "grit_score_metadata.json", metadata)
+    _atomic_json(commit_path, metadata)
+    success = output_dir / "_SUCCESS"
+    success.write_text(_sha256(commit_path) + "\n", encoding="utf-8")
+    return metadata
+
+
+def main(argv: list[str] | None = None) -> int:
```
vpraveen-nv: 🟠 **Major**: Two entrypoints for the same action: `dinov3 grit_score -e spec.yaml` (hydra schema + `monitor_status` + `status.json`) and this argparse `python -m ...cli --config grit.yaml` with a free-form dict. tao-data-services #40 (`recipes/grit_score.yaml`) invokes the argparse one, so in the real workflow TAO status logging, schema validation and `experiment.yaml` are bypassed and the 'native TAO action' is unused. Pick one: have the controller call `dinov3 grit_score -e ... grit_score.results_dir=...` and delete `main()` here (keep `run()` as the library function), or drop the TAO subtask. Two config surfaces will drift.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
```diff
@@ -0,0 +1,973 @@
...
+            crop_size = int(round(INPUT_SIZE * 0.9))
+            for corner, name in ((0, "view0"), (INPUT_SIZE - crop_size, "view1")):
+                crop = image.crop(
+                    (corner, corner, corner + crop_size, corner + crop_size)
+                ).resize((INPUT_SIZE, INPUT_SIZE), Image.Resampling.BICUBIC)
+                output[name] = self._tensor(crop)
+        return output
+
+
+def _build_backbone(base_spec: str | Path, checkpoint: str | Path) -> torch.nn.Module:
```
vpraveen-nv, line 619: 🟠 **Major**: `_build_backbone` re-implements `DinoV3PlModel._resolve_arch` + `_make_backbone`: re-reads `map_params`, hard-codes `{"mlp": Mlp, "swiglu": SwiGLUFusedFull}` (pl_model uses `_MLP_LAYERS`), and hard-codes `norm_layer/act_layer/qkv_bias`. Any change to training-side construction silently produces a GRIT backbone that no longer matches the checkpoint. Make `_make_backbone` (or a new `DinoV3PlModel.build_backbone(...)`) a classmethod and call it from here, next to the `load_backbone_weights` classmethod you already added.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
```diff
@@ -0,0 +1,973 @@
...
+    positive = logits.diagonal(dim1=1, dim2=2)
+    nll = torch.logsumexp(logits, dim=-1) - positive
+    count = max(
+        1, int(math.ceil(nll.shape[-1] * PATCH_RETRIEVAL_CVAR_FRACTION))
+    )
+    return nll.topk(count, dim=-1).values.mean(dim=-1)
+
+
+@torch.inference_mode()
+def extract_grit_scores(config: dict) -> pd.DataFrame:
```
vpraveen-nv, line 708: 🟠 **Major**: Nothing in `test_grit_score.py` runs the real extraction path: `extract_grit_scores` is only reached in a test that fails before the backbone is built, `_build_backbone` is tested with the ViT and `DinoV3PlModel` mocked out, and `_selected_layers`, `_pool_patches`, `_mapped_indices`, `_patch_retrieval`, `_ManifestImages` (file/tar/zip reads, view crops) and `forward_selected_layers` have no coverage. A CPU test with a tiny `DinoV3VisionTransformer` (embed_dim 32, depth 4, 2-3 PNGs, `workers=0`, `device=cpu`, `settling_k=1`, `view_ks=(1,)`) would exercise all of it in seconds and catch register-token slicing regressions. Please add it.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
```diff
@@ -0,0 +1,973 @@
...
+        observations.loc[rows, "dense_consensus"] = np.sqrt(
+            _rank(dense_values[rows]) * _rank(patch_retrieval[rows])
+        )
+    result = score_grit_frame(observations)
+    result.attrs["realized_neighbor_backends"] = sorted(realized_backends)
+    result.attrs["realized_neighbor_device"] = str(torch.device(neighbor_device))
+    return result
+
+
+class _ManifestImages(Dataset):
```
vpraveen-nv, line 571: 🟡 **Minor**: `_ManifestImages` duplicates `DinoV3Dataset` from #133 (file-URI parsing, tar/zip LRU handle cache, member index) but drops the fork-safety reopen (`_archive_owner_pid`) that #133 added. It's safe today only because the parent never opens an archive before workers fork; that's fragile. Reuse `DinoV3Dataset` (with `train=False` and a GRIT transform) or share the locator/archive helpers from `dataloader/dataset.py`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
```diff
@@ -0,0 +1,973 @@
...
+                device=compute_device,
+            )
+            for reference_start in range(0, len(reference), block_rows):
+                reference_stop = min(
+                    len(reference), reference_start + block_rows
+                )
+                reference_block = torch.from_numpy(
+                    reference_values[reference_start:reference_stop]
+                ).to(compute_device)
+                similarities = query_block @ reference_block.T
```
vpraveen-nv, line 135: 🟡 **Minor**: `_blockwise_neighbors` computes `query @ reference.T` and calls it 'exact cosine' but never L2-normalises, whereas `_faiss_neighbors` does (`faiss.normalize_L2`). The two backends only agree because callers pre-normalise. `derive_grit_observations` is a public export taking arbitrary arrays — either normalise inside `_blockwise_neighbors` too, or assert unit norm at the boundary so `auto` vs `faiss_exact` can never rank differently.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
```diff
@@ -0,0 +1,973 @@
...
+                "path",
+                "storage_type",
+                "member",
+                "task",
+                "role",
+            )
+            if name in available
+        ],
+    )
+    required = {"sample_id", "path", "storage_type", "task", "role", "embedding"}
```
vpraveen-nv: ❔ **Question**: `extract_grit_scores` hard-fails without an `embedding` column and copies it into the output, but GRIT never uses it — this is a DS retrieval concern leaking into the model-side scorer. Can the DS controller join embeddings by `sample_id` itself? If so, make `embedding` optional here (pass through if present) so the scorer works standalone.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
```diff
@@ -0,0 +1,973 @@
...
+            if layer_global is None:
+                layer_global = np.memmap(
+                    work_dir / "global_layers.f32",
+                    mode="w+",
+                    dtype=np.float32,
+                    shape=(len(manifest), *global_numpy.shape[1:]),
+                )
+            indices = batch["index"].numpy()
+            layer_global[indices] = global_numpy
+            dense_settling[indices] = _patch_settling(
```
vpraveen-nv, line 873: 🟡 **Minor**: Perf: the first pass computes pooled patch features at all four depths and `_patch_settling` for *every* manifest row, but only query rows are consumed (`dense_values[query_rows]`), and the second pass re-runs the canonical forward for queries whose final-layer CLS is already in `layer_global[:, 3]`. For a large reference pool that's ~2x wasted GPU time. Skip settling for reference rows and reuse the canonical CLS from pass one.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
```diff
@@ -0,0 +1,973 @@
...
+import torch.nn.functional as F
+from torch.utils.data import DataLoader, Dataset
+from torchvision.transforms import functional as TF
+
+from .grit import score_grit_frame
+
+
+IMAGE_MEAN = (0.485, 0.456, 0.406)
+IMAGE_STD = (0.229, 0.224, 0.225)
+INPUT_SIZE = 512
```
vpraveen-nv, line 29: ❔ **Question**: `INPUT_SIZE = 512` is hard-coded and every image is squashed to 512x512 regardless of `model.backbone.img_size` in the base spec (`_build_backbone` also passes `img_size=INPUT_SIZE`). I understand the 'frozen formula' argument, but please confirm this is intentional for non-square datasets, and surface it in `GRITScoreConfig` (even read-only/validated) so it appears in `experiment.yaml`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
```diff
@@ -0,0 +1,194 @@
...
+        value, sort_keys=True, separators=(",", ":"), ensure_ascii=True
+    ).encode("utf-8")
+    return "sha256:" + hashlib.sha256(encoded).hexdigest()
+
+
+def run(config: dict) -> dict:
+    """Execute scoring from a plain configuration dictionary."""
+    input_path = Path(config["input_parquet"]).resolve()
+    output_dir = Path(config["output_dir"]).resolve()
+    if (output_dir / "_SUCCESS").exists():
```
vpraveen-nv, line 62: 🟡 **Minor**: Two issues with the overwrite guard: (1) via the TAO subtask, `monitor_status` has already written `experiment.yaml` and `status.json` into this 'committed' directory before `run()` is entered, so it is mutated after all; (2) `RuntimeError` isn't in the exception list `monitor_status` catches, so no FAILURE status is emitted. Raise `FileExistsError` (→ 'File system error' status) and, for the TAO path, move the `_SUCCESS` check into `scripts/grit_score.py` before `_run_experiment`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
```diff
@@ -0,0 +1,194 @@
...
+                "available_scratch_bytes_at_start": result.attrs[
+                    "available_scratch_bytes_at_start"
+                ],
+                "patch_size": result.attrs["patch_size"],
+                "archive_cache_size": int(config.get("archive_cache_size", 8)),
+                "amp": bool(config.get("amp", True)),
+            }
+        )
+    commit_path = output_dir / "score_commit.json"
+    _atomic_json(output_dir / "grit_score_metadata.json", metadata)
```
vpraveen-nv: ⚪ **Nit**: `grit_score_metadata.json` and `score_commit.json` are byte-identical, and `_SUCCESS` is written non-atomically right after two atomic writes. Keep one metadata file and use `_atomic_json`/rename for `_SUCCESS` too.

nvidia_tao_pytorch/ssl/dinov3/scripts/grit_score.py
```diff
@@ -0,0 +1,50 @@
...
+    sizes = ("WORLD_SIZE", "LOCAL_WORLD_SIZE")
+    ranks = ("RANK", "LOCAL_RANK", "NODE_RANK")
+    if any(int(os.environ.get(name, "1")) != 1 for name in sizes) or any(
+        int(os.environ.get(name, "0")) != 0 for name in ranks
+    ):
+        raise ValueError("DINOv3 grit_score requires a single process on one node")
+    _run_experiment(cfg)
+
+
+@monitor_status(name="DINOv3", mode="grit_score")
```
vpraveen-nv, line 50: 🟡 **Minor**: TAO pattern is `@hydra_runner` + `@monitor_status` stacked on `main`. Splitting them so the topology check runs first is reasonable, but `mode="grit_score"` results-dir handling then happens in `_run_experiment` while `main` was called with an un-normalised cfg. Consider doing the env check in a small helper at the top of the `monitor_status`-decorated function (the FAILURE record is actually useful there), restoring the standard decorator stack.

nvidia_tao_pytorch/ssl/dinov3/model/vit.py
```diff
@@ -317,3 +317,31 @@ def forward(self, x, masks=None, keep_last_n_layers=None, keep_level: str = "chu
             "x_norm_patchtokens": x_norm[:, 1:],
             "masks": masks,
         }
+
```
vpraveen-nv, line 321: 🟡 **Minor**: `forward_selected_layers` duplicates the embedding + block loop from `forward(keep_last_n_layers=...)`. Cleaner: generalise `forward` with a `keep_layers: Iterable[int] | None` argument (keeping `keep_last_n_layers` as sugar) so there's one block loop and one register-token convention to maintain.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/__init__.py
```diff
@@ -2,3 +2,32 @@
...
+    "derive_grit_observations": (".grit_pipeline", "derive_grit_observations"),
+    "relative_layer_numbers": (".grit_pipeline", "relative_layer_numbers"),
+    "score_grit_frame": (".grit", "score_grit_frame"),
+}
+__all__ = list(_LAZY_EXPORTS)
+
+
+def __getattr__(name: str) -> Any:
+    """Load optional GRIT dependencies only when their APIs are requested."""
+    try:
```
vpraveen-nv: ⚪ **Nit**: The `TYPE_CHECKING` + `__getattr__` lazy-export dance is more machinery than the package needs; `grit.py` only imports numpy/pandas and `grit_pipeline` imports torch, which every consumer already has. Plain imports (or no re-exports) would be simpler.

tests/ssl_unit_test/dinov3/test_grit_score.py
```diff
@@ -0,0 +1,333 @@
...
+            "global_consensus": [0.2, 0.8],
+            "dense_consensus": [0.8, 0.2],
+        }
+    )
+    frame[column] = [invalid, invalid]
+    with pytest.raises(ValueError, match="finite real numbers"):
+        score_grit_frame(frame)
+
+
+def test_grit_backbone_reads_custom_attention_from_train_config(
```
vpraveen-nv, line 48: 🟡 **Minor**: This replaces both `...model.vit` and `...pl_model` in `sys.modules` and only asserts `use_custom_attention` was forwarded, so it can't detect a mismatch between `_build_backbone` and the real `DinoV3VisionTransformer` signature or `_MLP_LAYERS`. Once `_build_backbone` reuses `DinoV3PlModel`'s constructor, this can build a real tiny backbone instead.

nvidia_tao_pytorch/ssl/dinov3/scripts/grit_score.py
nnagrajrao: Fixed. The root supported-command table includes grit_score and the final repository static hook passes.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
nnagrajrao: Fixed. The controller calls dinov3 grit_score -e through the standard Hydra/monitor_status action. data_refinement.cli now supplies the library run() used by that native action; there is no second argparse entrypoint.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: Fixed. GRIT calls DinoV3PlModel.build_backbone and load_backbone_weights, so training and scoring share architecture selection and construction.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: Coverage was expanded with real tiny-backbone construction, file/archive reader tests, selected-layer/pooling/mapping/retrieval tests, runtime Parquet spies, and dummy-image GPU scoring smokes. The final PyTorch matrix is 246 passed with four reference-weight skips.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: Fixed. GRIT and manifest training share LocalImageReader, including fork/spawn ownership, archive member indexes, and bounded handle eviction.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: Fixed. Both Torch and FAISS paths consume the same robust finite, nonzero normalization boundary, and tests cover tiny/overflowing vectors plus backend parity.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: Fixed. embedding is optional for standalone GRIT; it is passed through by sample_id only when supplied for downstream DS retrieval.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: This optimization was intentionally not mixed into the correctness change: the implementation remains disk-backed and memory-bounded, and computes the frozen observation formula identically for all rows. Production-scale throughput remains an explicit residual benchmark item.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: Confirmed intentional. input_size is now a validated GRITScoreConfig field fixed to 512 and is recorded in experiment/commit metadata; the resize policy is documented for non-square inputs.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
nnagrajrao: Fixed. The native entrypoint checks topology and _SUCCESS before monitor_status can mutate the committed directory; output payloads and _SUCCESS use atomic replacement.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
nnagrajrao: Fixed. One score_commit.json is authoritative and _SUCCESS is atomically published with its digest; the duplicate metadata file was removed.

nvidia_tao_pytorch/ssl/dinov3/scripts/grit_score.py
nnagrajrao: The final native path uses Hydra plus monitor_status, but performs topology/committed-output preflight before entering monitor_status so invalid distributed launches cannot write shared status artifacts. This is deliberate and covered by entrypoint tests.

nvidia_tao_pytorch/ssl/dinov3/model/vit.py
nnagrajrao: The duplicate block loop was removed into the shared _forward_blocks helper; forward_selected_layers and forward now use the same token/register convention and block execution.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/__init__.py
nnagrajrao: The lazy exports are retained to keep importing the lightweight score-formula package from eagerly importing Lightning/model dependencies. This is an import-isolation choice rather than a runtime correctness requirement.

tests/ssl_unit_test/dinov3/test_grit_score.py
nnagrajrao: Fixed. Tests build the real tiny DinoV3 backbone through DinoV3PlModel.build_backbone and exercise the actual constructor contract rather than replacing model modules in sys.modules.

nvidia_tao_pytorch/ssl/dinov3/scripts/grit_score.py
vpraveen-nv: Verified at c22f3b8 — `README.md:90` lists `grit_score` and `update_readme_supported_commands.py --check` passes. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
vpraveen-nv: Verified at c22f3b8 — `cli.py` is library-only (`run(config)`), and DS #40's `recipes/grit_score.yaml` / controller use `[dinov3, grit_score, -e, {score_config}]`. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
vpraveen-nv: Verified — `grit_pipeline._build_backbone` calls `DinoV3PlModel.build_backbone(...)` + `load_backbone_weights`, and `_make_backbone` delegates to the same classmethod. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
vpraveen-nv: Partially resolved. `test_real_cpu_extraction_reads_file_tar_zip_and_excludes_registers` does exercise `_ManifestImages` + `_selected_layers` with a real tiny ViT — good. But nothing in this repo runs `extract_grit_scores` end to end: every test that reaches it either fails in validation or monkeypatches `_build_backbone` to `pytest.fail`, so the two DataLoader passes, the memmaps, `_patch_retrieval`/`_mapped_indices`, the `embedding` join and the `result.attrs` bookkeeping have zero coverage here. One CPU test that monkeypatches `_build_backbone` to return a tiny `DinoV3VisionTransformer(img_size=512, patch_size=16, embed_dim=32, depth=4, register_tokens=2)` with ~6 PNGs, `workers=0`, `settling_k=1`, `view_ks=[1]` would run in seconds. The GPU scoring smokes you mention live in the DS cross-repo smoke, not in `tests/ssl_unit_test/dinov3`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
vpraveen-nv: Verified — `_ManifestImages` now wraps `LocalImageReader` (`dataloader/local_image.py`), which drops handles in `__getstate__` and reopens on `_owner_pid` mismatch. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
vpraveen-nv: Verified — `_blockwise_neighbors` normalises both query and reference blocks with `F.normalize`, FAISS path uses `normalize_L2`; `test_blockwise_neighbors_match_full_matrix` covers the torch path. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
vpraveen-nv: Verified — `embedding` is read only `if "embedding" in available` and joined by `sample_id`. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
vpraveen-nv: Understood — accepting the deferral. Please track it (settling for reference rows + canonical CLS re-forward are still ~2x on large reference pools) so it doesn't get lost after merge.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
vpraveen-nv: Verified — `GRITScoreConfig.input_size = INT_FIELD(512, valid_options="512")`, enforced in `extract_grit_scores`, recorded in `score_commit.json`; tao-core #38 carries the same field and asserts `enum == [512]`. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
vpraveen-nv: Half resolved. The `_SUCCESS`/topology preflight now runs in `main()` before `monitor_status` — good. But `run()` still raises `RuntimeError` (attestation mismatch at `cli.py:73`, repeat `_SUCCESS` check at `cli.py:77`) inside `_run_experiment`, and `monitor_status` does not catch `RuntimeError`, so a controller-lock mismatch exits with a bare traceback and `status.json` never records FAILURE. Raise a type the decorator maps (`ValueError` → 'Configuration error' or `PermissionError` → 'File system error'), or extend the decorator.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
vpraveen-nv: Verified — single `score_commit.json`, `_SUCCESS` written via rename with the commit digest. Resolved, thanks. Small nit left inline about reusing the fsync'd writer from `refinement_attestation.py`.

nvidia_tao_pytorch/ssl/dinov3/scripts/grit_score.py
vpraveen-nv: Fine — the rationale (no shared artifacts before topology validation) is sound and `test_tao_grit_rejects_distributed_workers_before_publication` covers it. Resolved.

nvidia_tao_pytorch/ssl/dinov3/model/vit.py
vpraveen-nv: Verified — `_forward_blocks` is the single block loop; `forward` and `forward_selected_layers` share it and the append-at-end register convention. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/__init__.py
vpraveen-nv: The head actually has plain imports now (`from .grit import GRIT_FORMULA_VERSION, score_grit_frame`, no `TYPE_CHECKING`/`__getattr__`), which is what I asked for — resolved regardless of the reply text.

tests/ssl_unit_test/dinov3/test_grit_score.py
vpraveen-nv: Verified — `test_grit_backbone_reads_custom_attention_from_train_config` builds a real `vit_s` via `DinoV3PlModel.build_backbone`; the only remaining `sys.modules` patch is the fake `faiss`. Resolved, thanks.

nvidia_tao_pytorch/ssl/dinov3/utils/refinement_attestation.py
```diff
@@ -0,0 +1,162 @@
...
+
+def canonical_digest(value: dict) -> str:
+    """Hash a JSON object using the controller's canonical encoding."""
+    encoded = json.dumps(
+        value, sort_keys=True, separators=(",", ":"), ensure_ascii=True
+    ).encode("utf-8")
+    return "sha256:" + hashlib.sha256(encoded).hexdigest()
+
+
+def implementation_files(subtask: str) -> dict[str, Path]:
```
vpraveen-nv, line 32: 🟠 **Major**: This closure (directory list + `required` set + `native.<rel>` keys + `canonical_digest`) is duplicated verbatim in DS #40 `_native_implementation_files` (controller side). Any drift — adding a directory here, or DS adding one there — makes every DEFT run fail closed with 'does not match the controller lock', and the only cross-check (`test_native_lock_covers_dinov3_and_inherited_nvdinov2_runtime` in DS) is `pytest.skip`ped whenever tao-pytorch isn't installed, i.e. in DS CI. Since DS `preflight` already imports `nvidia_tao_pytorch.ssl.dinov3.scripts.grit_score`, please make this module the single owner: have DS import `implementation_files`/`native_attestation` from here (or shell out to `python -m nvidia_tao_pytorch.ssl.dinov3.utils.refinement_attestation <subtask>` emitting JSON) instead of re-deriving the list. At minimum add a cross-reference comment in both files and a non-skipping parity test in one of them.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
```diff
@@ -0,0 +1,178 @@
...
+def _canonical_digest(value: dict) -> str:
+    encoded = json.dumps(
+        value, sort_keys=True, separators=(",", ":"), ensure_ascii=True
+    ).encode("utf-8")
+    return "sha256:" + hashlib.sha256(encoded).hexdigest()
+
+
+def run(config: dict) -> dict:
+    """Execute scoring from a plain configuration dictionary."""
+    attestation = verify_native_attestation("grit_score")
```
vpraveen-nv, line 59: 🟡 **Minor**: `verify_native_attestation` raises `RuntimeError` here, inside `_run_experiment` (`@monitor_status`). The decorator maps `ValueError/TypeError`, the `OSError` family, `NotImplementedError` and the Hydra/marshmallow errors, but not `RuntimeError`, so a controller-lock mismatch produces a bare traceback and `status.json` stays at STARTED with no FAILURE record for the DS controller to read. Same for the `_SUCCESS` re-check on line 77. Raise a mapped type (or subclass one) so the failure is visible in TAO status.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
```diff
@@ -0,0 +1,178 @@
...
+    if path.is_file():
+        return path
+    for name in ("model.safetensors", "pytorch_model.bin", "model.pth"):
+        candidate = path / name
+        if candidate.is_file():
+            return candidate
+    raise ValueError(f"Cannot fingerprint DINOv3 checkpoint: {path}")
+
+
+def _atomic_json(path: Path, value: dict) -> None:
```
vpraveen-nv: ⚪ **Nit**: `_atomic_json`/`_atomic_text` use a fixed `<name>.tmp` sibling and no `fsync`, while `publish_native_attestation` in the same PR uses `NamedTemporaryFile` + `fsync` + directory `fsync`. Single-process is enforced so the race is moot, but durability of `score_commit.json`/`_SUCCESS` on a crash right after `replace()` isn't. Consider exposing the attestation writer as a shared helper and using it here (and for `grit_scores.tmp.parquet`).

nvidia_tao_pytorch/ssl/dinov3/scripts/train.py
```diff
@@ -132,6 +135,7 @@ def run_experiment(experiment_config, key):
 def main(cfg: ExperimentConfig) -> None:
     """Run the training process."""
     publish_runtime_spec(cfg, cfg.results_dir)
+    publish_native_attestation("train", cfg.results_dir)
```
vpraveen-nv: ❔ **Question**: `publish_native_attestation("train", ...)` now runs for every ordinary `dinov3 train` (not just DEFT): it hashes ~145 files (cheap) but also writes `train_implementation_audit.json` into every user's results dir and hard-fails with `RuntimeError` if any closure directory is missing from the installed package. Is that intended for non-DEFT users, or should it be gated on the `TAO_REFINEMENT_*` env (or `dataset.train_manifest`) being set?

nvidia_tao_pytorch/config/dinov3/default_config.py
```diff
@@ -686,6 +688,10 @@ class DINOv3ConvertConfig:
 class ExperimentConfig(CommonExperimentConfig):
     """DINOv3 experiment config."""
 
+    grit_score: GRITScoreConfig = DATACLASS_FIELD(
```
vpraveen-nv: ⚪ **Nit**: Declaring `grit_score` before `model` makes `dinov3 default_specs` emit the GRIT block ahead of `model`/`dataset`/`train`. Please move it after `convert` (and mirror in tao-core #38, which has the same ordering) so generated specs keep the conventional layout.

vpraveen-nv (commented):
Thanks for the restack — I re-read the full head (c22f3b8). Verified: `data_refinement/cli.py` is library-only and DS #40 invokes `dinov3 grit_score -e` through Hydra + `monitor_status`; `_build_backbone` goes through `DinoV3PlModel.build_backbone` + `load_backbone_weights`; `_ManifestImages` reuses `LocalImageReader` with `_owner_pid` reopen; both neighbour backends normalise internally; `embedding` is optional; `input_size` is a validated `GRITScoreConfig` field mirrored in tao-core #38; `_SUCCESS` is checked before `monitor_status` and published via rename; README table and `_forward_blocks` refactor check out. 13 of 15 earlier threads resolved, 2 partial — replied inline.

The new `verify_native_attestation` gate matches DS #40's controller-side closure key-for-key today, but the closure definition is now hand-duplicated across two repos and the only parity test skips when tao-pytorch is absent (i.e. in DS CI) — that's my one **major**. Remaining: the end-to-end `extract_grit_scores` path is still never executed in this repo's tests, and a `RuntimeError` from the attestation gate inside `_run_experiment` bypasses `monitor_status` FAILURE reporting. Approve once the CPU end-to-end test is added; the closure-ownership item can be a fast follow-up if we agree where it lives.

nvidia_tao_pytorch/ssl/dinov3/utils/refinement_attestation.py
nnagrajrao: PyTorch's refinement_attestation module is now the single owner. DS imports its implementation closure instead of maintaining a copy; a closure_version is included in locks/audits and the mandatory cross-repo smoke checks the handshake.

Revision: `1b3720ae`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
nnagrajrao: Attestation and committed-output errors now use ValueError, which monitor_status maps to a TAO FAILURE/configuration-error record. No shared decorator exception policy was broadened.

Revision: `1b3720ae`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
nnagrajrao: GRIT JSON/text/Parquet publication now uses the shared DINOv3 atomic helper, including unique temporary paths and file/directory fsync before the success marker is published.

Revision: `1b3720ae`.

nvidia_tao_pytorch/ssl/dinov3/scripts/train.py
nnagrajrao: Gated training attestation on manifest/DEFT execution. Ordinary directory-backed DINOv3 training no longer emits this audit or requires the DEFT implementation closure.

Revision: `1b3720ae`.

nvidia_tao_pytorch/config/dinov3/default_config.py
nnagrajrao: Moved grit_score after convert in both runtime and Core ExperimentConfig. Schema parity is checked in the container suite.

Revision: `1b3720ae`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: Added test_extract_grit_scores_cpu_end_to_end using a real tiny ViT and generated PNGs. It drives both DataLoader passes, scratch memmaps, patch retrieval/index mapping, embedding join, and result attrs. It passes in the final PyTorch source suite.

Revision: `1b3720ae`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/grit_pipeline.py
nnagrajrao: Still open / partially addressed. This optimization remains deferred as agreed: skip settling for reference-only rows and reuse first-pass canonical CLS rather than re-forwarding queries. It is recorded in the local TAO-2529 follow-up ledger; no throughput improvement is claimed.

Revision: `1b3720ae`.

nvidia_tao_pytorch/ssl/dinov3/data_refinement/cli.py
nnagrajrao: Completed the remaining half: attestation and repeated committed-output validation now raise ValueError so the existing TAO status decorator records FAILURE. The early pre-monitor success guard remains.

Revision: `1b3720ae`.

vpraveen-nv (approved):
Verified all five round-2 items at 1b3720a: `refinement_attestation.implementation_files()` is now the single closure owner with `CLOSURE_VERSION` and DS #40 imports it (no more hand-copied directory list); attestation/`_SUCCESS` failures raise `ValueError` so `monitor_status` records FAILURE; `cli.py` uses the shared atomic helpers; `publish_native_attestation("train")` is gated on `train_manifest`/`TAO_REFINEMENT_*`; `grit_score` follows `convert` in both schema copies and the README table check passes. Static checks/DCO green. Approving.

Tiny follow-ups: switch the pre-`monitor_status` `_SUCCESS` guard in `scripts/grit_score.py` to `ValueError` for consistency; renumber README sections after the stack lands.

Issue #41, opened 2025-09-12
[QST] Tagged releases?


weiji14:
**What is your question?**

Would it be possible to tag the v6.0.0 and v6.25.7 here on GitHub? I only see v5.2.0 and older versions on https://github.com/NVIDIA/tao_pytorch_backend/tags. Asking so that it is easier to install from source because the current PyPI wheels (e.g. https://pypi.org/project/nvidia-tao-pytorch/6.25.7.0.2/#files) are for Python 3.12 only.

jakirkham:
Thanks Wei! 🙏

Missed this issue when writing up issue: https://github.com/NVIDIA/tao_pytorch_backend/issues/51

It does seem like there are commit messages that reference versioned releases. They just are not tagged. Will ask about this

weiji14:
Gonna close this since v7.0.0 was tagged properly - https://github.com/NVIDIA-TAO/tao-pytorch/releases/tag/7.0.0.

I've also pushed an update for the package on conda-forge now - https://github.com/conda-forge/nvidia-tao-pytorch-feedstock/pull/7

PR #136, opened 2026-09-13
[TAO-2591][Feature] Add Sparse4D mixed 2D/3D co-training

yizwang-nv:
JIRA: [TAO-2591](https://jirasw.nvidia.com/browse/TAO-2591)

## Summary

- add route-synchronized mixed 2D/3D co-training for Sparse4D
- add Loose-to-Tight matching, checkpoint validation, pseudo-label losses, and the single-view auxiliary head
- add lazy multi-PKL indexing/resampling and self-contained LTT, RT-DETR, and SV2D artifact tools
- harden temporal state, distributed loss accounting, training resume behavior, export contracts, and SV2D cache/image dimension validation
- add comprehensive self-contained unit, contract, and trainer integration coverage

## Compatibility

- all new co-training paths are disabled by default
- existing 3D-only training, evaluation, inference, and export behavior remains available
- this change does not modify dependencies, Dockerfiles, `.gitmodules`, or the `tao-core` gitlink
- the public-main optimizer-step resume fix is preserved in this branch

## Validation

- `pytest -q tests/cv_unit_test/sparse4d`: **150 passed, 1 skipped**
- Pylint on all changed package sources: **10.00/10**
- license-header check: passed
- pydocstyle: passed
- Flake8: passed
- `git diff --check`: passed
- one-GPU train/resume and two-GPU synchronized-route training/validation were exercised in the released TAO 7.2 PyTorch container on the source-equivalent implementation
- cross-repository producer/consumer smoke passed with tao-data-services #41 for LTT, RT-DETR, lazy loading, HDF5 image loading, and SV2D canonical resize

## Companion PRs and landing order

- [NVIDIA-TAO/tao-core#39](https://github.com/NVIDIA-TAO/tao-core/pull/39)
- [NVIDIA-TAO/tao-data-services#41](https://github.com/NVIDIA-TAO/tao-data-services/pull/41)

Preferred merge order: this runtime PR, then tao-core #39, then tao-data-services #41. Reviews can proceed in parallel.


nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
```diff
@@ -327,45 +1152,53 @@ def forward(self,
...
         """
         num_classes = pred.size(1)
-        valid_mask = (target >= 0) & (target < num_classes)  # ignore negative/out-of-range
-        if not valid_mask.any():
-            # No valid samples, return zero loss
-            return pred.new_tensor(0.0)
+        # Sparse4D follows the MMDetection sigmoid-focal convention: target C is
+        # background (an all-zero one-hot row), while negative and >C labels are
+        # ignored. The previous TAO implementation dropped every background query.
+        valid_mask = (target >= 0) & (target <= num_classes)
```
vpraveen-nv: **[major]** This changes the classification loss for **every** existing 3D run, not just co-training. `main` uses `(target >= 0) & (target < num_classes)`, which drops all background queries (`SparseBox3DTarget` fills unmatched queries with `num_cls`), so negatives were never penalised. Including them (`target <= num_classes`) is almost certainly the right fix — but it materially changes loss scale and training dynamics, and it contradicts the description's "existing 3D-only train … unchanged". Please (a) call it out explicitly, (b) land it as its own commit/PR with a before/after mAP on the release checkpoint, and (c) confirm the released 3D specs/LR still converge. If reproducibility of old runs matters, gate the legacy behaviour behind a flag.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_head.py
```diff
@@ -327,45 +341,72 @@ def graph_model(
...
+    def _reset_at_data_boundary(self, metas):
+        """Reset all temporal caches when a group or scene changes."""
+        group_indices = self._get_data_indices(metas, "group_idx")
+        scene_indices = self._get_data_indices(metas, "scene_idx")
+        previous_groups, previous_scenes = self.instance_bank.get_data_indices()
+
+        boundary = self._indices_changed(previous_groups, group_indices)
+        boundary = boundary or self._indices_changed(previous_scenes, scene_indices)
+        if boundary:
+            self.instance_bank.reset_temporal_state()
```
vpraveen-nv: **[major]** `_reset_at_data_boundary` runs unconditionally (`main` gated this on `use_temporal_align`) and, when *any* sample in the batch changes `group_idx`/`scene_idx`, wipes the **entire** batch's recurrent state (`cached_feature`, `cached_anchor`, `instance_id`, `confidence`, …) plus `sampler.dn_metas`. `main` only reset the GT-index mapping for the affected samples and relied on `InstanceBank.mask` (per-sample `|dt| <= max_time_interval`) to stop propagation for that sample. With `GroupInBatchSampler` the per-slot buffers empty at different times (skip_prob=0.5, variable group sizes), so one slot rolling to a new group now truncates the temporal window for the other `batch_size-1` samples — a training-dynamics change for the existing 3D path.

Suggest keeping the per-sample mapping reset from `main`, and only doing a batch-wide `reset_temporal_state()` when **all** samples cross a boundary (which is exactly the `scene_switch_iters` co-training case), or gating the new behaviour behind a config flag.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
```diff
@@ -242,27 +608,493 @@ def forward(self, raw_model_outs, data, feature_maps=None):
...
+            tensors.extend(value for value in depths if torch.is_tensor(value))
+        if not tensors:
+            return torch.zeros(())
+        return sum(
+            torch.nan_to_num(tensor, nan=0.0, posinf=0.0, neginf=0.0).sum() * 0.0
+            for tensor in tensors
+        )
+
+    @staticmethod
+    def _finite_losses(output, zero_ref=None):
```
vpraveen-nv: **[major]** `_finite_losses` (plus the `cls.clamp(-50, 50)` and `nan_to_num(...).sum()*0.0` fallbacks around L446–455 / L580–596) turns a NaN/Inf loss into a zero with only a `warnings.warn`, on the default 3D path. Previously a NaN loss propagated and was visible immediately (and Lightning's anomaly detection / grad scaler would react). Silently continuing with zeroed losses can hide a diverging run for hours, and `warnings.warn` is de-duplicated by Python's filter so it fires once per location.

Please either gate this behind the existing opt-in `train.scrub_nan_gradients` (or a new `train.zero_nonfinite_losses`) so the default path fails loudly as before, or at minimum route it through `status_logging` with a running counter and raise after N consecutive non-finite steps.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_pl_model.py
```diff
@@ -29,7 +36,37 @@ def __init__(self, experiment_spec, export=False):
...
+                bool(ltt_config.get("enable", False)),
+                bool(ltt_config.get("pseudo_enable", False)),
+            )
+        )
+        route_specific_loss = bool(
+            ltt_pseudo_loss or sv_aux_config.get("enable", False)
+        )
+        # A configured sampler route is not proof of equal runtime branching.
+        sync_positive_counts = not route_specific_loss
+        self.criterion = SetCriterion(
```
vpraveen-nv: **[minor]** `SetCriterion` is constructed for every entrypoint (`evaluate.py`, `inference.py`, `export.py`, `quantize.py` all build `Sparse4DPlModel(cfg)`), and with `loose_to_tight.enable=true` its `__init__` now does `LooseToTightMLP.load(mlp_ckpt)` and raises `FileNotFoundError` if the training-only MLP checkpoint is absent. So a model trained with LTT can't be evaluated/exported from the same spec on a box without `mlp_ckpt`. Suggest building the LTT/SV-aux pieces lazily on first `training_step`, or skipping them when `export=True` / not in the fit stage.

nvidia_tao_pytorch/cv/sparse4d/dataloader/transforms.py
```diff
@@ -146,8 +185,16 @@ def __call__(self, results: Dict) -> Dict:
 
             depths = np.stack(depths, axis=-1)
         depths = depths.astype(np.float32)
-        depths /= 1000.  # Convert from mm to m
-        depths = np.clip(depths, 0.1, self.max_depth)
+        # Missing maps use a negative sentinel. Preserve invalid pixels while
+        # converting valid millimetre values to metres; clipping the entire
+        # array would otherwise turn missing supervision into a fake 0.1 m
+        # target.
+        valid_depth = np.logical_and(np.isfinite(depths), depths > 0.0)
```
vpraveen-nv: **[minor]** Good catch — `main`'s `np.clip(depths/1000, 0.1, max_depth)` turned the `-1` missing-map sentinel into a fake 0.1 m supervision target. But this changes `loss_dense_depth` for existing 3D training with missing depth maps and is unrelated to co-training. Please mention it in the description and ideally split it into its own commit so the depth-loss change can be bisected independently.

nvidia_tao_pytorch/cv/sparse4d/dataloader/callbacks.py
```diff
@@ -11,27 +11,22 @@
...
 
-    def on_train_batch_start(self, trainer, pl_module, batch, batch_idx):
-        """Check if we should resample at this step."""
-        global_step = trainer.global_step
-        if global_step == 0 or global_step % self.num_iters_per_epoch != 0:
-            return
-
-        current_epoch = global_step // self.num_iters_per_epoch
-
+    def on_train_epoch_end(self, trainer, pl_module):
```
vpraveen-nv: **[minor]** Moving resampling from `on_train_batch_start` to `on_train_epoch_end` together with `reload_dataloaders_every_n_epochs=1` (train.py ~L121) and manually calling `iterator._shutdown_workers()` is a behavioural rewrite of the existing `pkl_sample_size` feature, not a co-training change. Two things to double-check: (1) `GroupInBatchSampler` is rebuilt each epoch with the same `train.seed`, so the scene-permutation sequence repeats every epoch — consider mixing `trainer.current_epoch` into the seed; (2) `_iterator` / `_shutdown_workers` are private DataLoader attributes; a comment pinning the torch version this was verified on would help the next upgrade.

nvidia_tao_pytorch/cv/sparse4d/dataloader/dataset.py
```diff
@@ -513,26 +606,50 @@ def load_annotations_lazy(self, ann_file):
...
         self._full_frame_index = frame_index
 
         self._pkl_cam_counts = {}
-        if self.pkl_sample_size > 0 and self.pkl_cam_counts_path:
-            if osp.exists(self.pkl_cam_counts_path):
+        if self.pkl_sample_size > 0:
+            if self.pkl_cam_counts_path:
+                if not osp.exists(self.pkl_cam_counts_path):
+                    raise FileNotFoundError(
+                        "Configured pkl_cam_counts_path does not exist: "
```
vpraveen-nv: **[minor]** Backward-compat change: `main` logged a warning for a missing `pkl_cam_counts_path` and continued; this now raises `FileNotFoundError`, raises `ValueError` if the lazy index has no embedded `pkl_cam_counts`, and (~L150) raises if `pkl_sample_size > 0 and not lazy_load`. Existing specs relying on warn-and-continue will fail at startup. Fine if intentional — please note it in the description / release notes. Also, the error text here points users at `python -m nvidia_tao_pytorch.cv.sparse4d.tools.build_lazy_index`; if the producers consolidate into tao-ds (see summary) this should name `annotations sparse4d_prepare` instead.

nvidia_tao_pytorch/cv/sparse4d/dataloader/sampler.py
```diff
@@ -130,21 +155,115 @@ def _sync_random_seed(self, seed=None):
...
+                        "GroupInBatchSampler loaded synthetic scene weights from %s",
+                        sidecar_path,
+                    )
+
+        self.scene_switch_iters = int(
+            getattr(self.dataset, "scene_switch_iters", 0) or 0
+        )
+        if self.sync_route and self.scene_switch_iters <= 0:
+            logging.warning(
+                "sync_route is enabled with scene_switch_iters <= 0; ranks "
```
vpraveen-nv: **[minor]** With `sync_route=True` and `scene_switch_iters <= 0` this only warns, yet the whole route-lockstep guarantee (each world-size block drawn from one route, ranks slicing `rank::world_size`) depends on every rank advancing scenes at the same step. Without a fixed cadence, ranks drift after the first differing group length and `_sync_cotrain_loss_keys` starts zero-filling losses every step (logged up to 20 times, then silent). Suggest raising `ValueError` here when `world_size > 1`, or auto-deriving a cadence.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
```diff
@@ -242,27 +608,493 @@ def forward(self, raw_model_outs, data, feature_maps=None):
...
+            metas = data.get("img_metas")
+            if isinstance(metas, (list, tuple)):
+                values = [meta.get(self.ltt_has3dgt_key) for meta in metas]
+        if values is None:
+            return False
+        if isinstance(values, torch.Tensor):
+            values = values.detach().cpu().reshape(-1).tolist()
+        elif not isinstance(values, (list, tuple)):
+            values = [values]
+        return bool(values) and all(
```
vpraveen-nv: **[question]** `_is_real_batch` requires **all** samples to be marked `has_3d_gt=False`. A mixed batch (possible with `same_scene_in_batch=False`, or when a real scene contains a few frames that do carry 3D boxes) silently falls through to the 3D loss path: the real samples contribute only background targets and the pseudo/SV losses are skipped with no log. Intended? If mixed batches are unsupported, assert homogeneity and raise so misconfiguration is visible.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_pl_model.py
```diff
@@ -63,28 +102,92 @@ def training_step(self, batch, batch_idx):
...
+    def _sync_cotrain_loss_keys(self, losses, zero_ref):
+        """Canonicalize heterogeneous route loss dictionaries across DDP ranks."""
+        if not (dist.is_available() and dist.is_initialized()):
+            return losses
+        world_size = dist.get_world_size()
+        if world_size == 1:
+            return losses
+
+        gathered = [None] * world_size
+        dist.all_gather_object(gathered, sorted(losses))
```
vpraveen-nv: **[suggestion]** `dist.all_gather_object(gathered, sorted(losses))` pickles/broadcasts a Python list on every training step whenever `pseudo_enable` or `sv_aux_head.enable` is set. The loss-key set is a deterministic function of route × config, so you could compute the canonical key set once in `__init__` and always emit every key (zero-filled), removing the per-step collective and the desync log path. That also gives loggers a stable loss-dict shape.

nvidia_tao_pytorch/cv/sparse4d/model/instance_bank.py
```diff
@@ -162,25 +179,37 @@ def reset_gt_index_mapping_by_data_indices(self, reset_flags):
...
+        if self.cached_anchor is not None and batch_size == self.cached_anchor.shape[0]:
             history_time = self.metas["timestamp"]
             time_interval = metas["timestamp"] - history_time
             time_interval = time_interval.to(dtype=instance_feature.dtype)
             self.mask = torch.abs(time_interval) <= self.max_time_interval
 
+            # A long gap or clip-boundary timestamp reset can otherwise carry a
+            # poisoned recurrent feature into every later clip. Keep IDs globally
+            # unique, but discard all temporal tensors when explicitly requested.
+            if self.reset_on_time_gap and not bool(torch.all(self.mask).item()):
```
vpraveen-nv: **[question]** Two questions on reset semantics: (1) the `else` branch (~L256) now calls `reset_temporal_state()` instead of `reset()`, so `prev_id` is no longer zeroed when the cache is missing / batch size changes — for test-time tracking that means instance IDs keep counting across scenes rather than restarting at 0. Does the NVSchema/HOTA path tolerate that? (2) `reset_on_time_gap` resets the **whole batch** when any single sample's `mask` is False, even though `mask` is already applied per-sample at ~L301–306. Should this be `torch.all(~mask)`, or a per-sample zeroing?

nvidia_tao_pytorch/cv/sparse4d/utils/onnx_export.py
```diff
@@ -15,6 +15,18 @@
 from nvidia_tao_pytorch.cv.sparse4d.model.ops.deformable_aggregation import DeformableAggregationFunction, feature_maps_format
 
 
+def sanitize_sampling_locations_for_export(points_2d):
```
vpraveen-nv: **[question]** This inserts `torch.where(valid, pts, 0)` into the exported graph for **all** Sparse4D exports (the exporter's static method now delegates here), so the ONNX from an unchanged 3D checkpoint differs from what `main` produces — contradicting "export unchanged". Has the resulting ONNX been run through tao-deploy `gen_trt_engine` + inference with the current MSDA TRT plugin to confirm numerical parity? Worth stating in the docstring that this relies on the plugin rejecting exactly-0 samples (`loc <= 0` returns today).

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d.py
```diff
@@ -42,22 +46,36 @@ def __init__(
         self.neck_config = self.model_config["neck"]
         self.head_config = self.model_config["head"]
         self.depth_branch_config = self.model_config["depth_branch"]
+        self.cotrain_param_touch = bool(
```
vpraveen-nv: **[nit]** `self.cotrain_param_touch` is assigned here but never read inside `Sparse4D` (the touch lives in `Sparse4DPlModel._parameter_touch_zero`). Remove it, or drop the comment block referencing timm `checkpoint_seq` if it no longer applies.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
```diff
@@ -135,20 +420,40 @@ def forward(self, raw_model_outs, data, feature_maps=None):
...
+                        instance_id_target,
+                        data,
+                        num_pos,
+                    )
+                )
+
             cls = cls.flatten(end_dim=1)
             cls_target = cls_target.flatten(end_dim=1)
-            cls_loss = self.loss_cls(cls, cls_target, avg_factor=num_pos)
+            cls_input = cls.clamp(min=-50.0, max=50.0)
```
vpraveen-nv: **[nit]** The `clamp(-50, 50)` on classification logits also applies on the default 3D path. Numerically a no-op for healthy logits, but it silently zeroes gradients for any query whose logit exceeds the bound — another small deviation from `main` for existing runs. Fold it into the same opt-in stability flag as the non-finite handling, or document why always-on is safe.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
vpraveen-nv: **[major]** This changes the classification loss for **every** existing 3D run, not just co-training. `main` uses `(target >= 0) & (target < num_classes)`, which drops all background queries (`SparseBox3DTarget` fills unmatched queries with `num_cls`), so negatives were never penalised. Including them (`target <= num_classes`) is almost certainly the right fix — but it materially changes loss scale and training dynamics, and it contradicts the description's "existing 3D-only train … unchanged". Not asking you to split it out — please (a) call it out explicitly in the description, and (b) post a before/after mAP on the release 3D checkpoint from this branch so we know the released specs/LR still converge. If reproducibility of old runs matters, a flag for the legacy behaviour would be cheap insurance, but I won't block on that.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_head.py
vpraveen-nv: **[major]** `_reset_at_data_boundary` runs unconditionally (`main` gated this on `use_temporal_align`) and, when *any* sample in the batch changes `group_idx`/`scene_idx`, wipes the **entire** batch's recurrent state (`cached_feature`, `cached_anchor`, `instance_id`, `confidence`, …) plus `sampler.dn_metas`. `main` only reset the GT-index mapping for the affected samples and relied on `InstanceBank.mask` (per-sample `|dt| <= max_time_interval`) to stop propagation for that sample. With `GroupInBatchSampler` the per-slot buffers empty at different times (skip_prob=0.5, variable group sizes), so one slot rolling to a new group now truncates the temporal window for the other `batch_size-1` samples — a training-dynamics change for the existing 3D path.

Suggest keeping the per-sample mapping reset from `main`, and only doing a batch-wide `reset_temporal_state()` when **all** samples cross a boundary (which is exactly the `scene_switch_iters` co-training case), or gating the new behaviour behind a config flag.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
vpraveen-nv: **[major]** `_finite_losses` (plus the `cls.clamp(-50, 50)` and `nan_to_num(...).sum()*0.0` fallbacks around L446–455 / L580–596) turns a NaN/Inf loss into a zero with only a `warnings.warn`, on the default 3D path. Previously a NaN loss propagated and was visible immediately (and Lightning's anomaly detection / grad scaler would react). Silently continuing with zeroed losses can hide a diverging run for hours, and `warnings.warn` is de-duplicated by Python's filter so it fires once per location.

Please either gate this behind the existing opt-in `train.scrub_nan_gradients` (or a new `train.zero_nonfinite_losses`) so the default path fails loudly as before, or at minimum route it through `status_logging` with a running counter and raise after N consecutive non-finite steps.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_pl_model.py
vpraveen-nv: **[minor]** `SetCriterion` is constructed for every entrypoint (`evaluate.py`, `inference.py`, `export.py`, `quantize.py` all build `Sparse4DPlModel(cfg)`), and with `loose_to_tight.enable=true` its `__init__` now does `LooseToTightMLP.load(mlp_ckpt)` and raises `FileNotFoundError` if the training-only MLP checkpoint is absent. So a model trained with LTT can't be evaluated/exported from the same spec on a box without `mlp_ckpt`. Suggest building the LTT/SV-aux pieces lazily on first `training_step`, or skipping them when `export=True` / not in the fit stage.

nvidia_tao_pytorch/cv/sparse4d/dataloader/transforms.py
vpraveen-nv: **[minor]** Good catch — `main`'s `np.clip(depths/1000, 0.1, max_depth)` turned the `-1` missing-map sentinel into a fake 0.1 m supervision target. But this changes `loss_dense_depth` for existing 3D training with missing depth maps and is unrelated to co-training. Please mention it in the description (and in the CHANGELOG line for this feature) so anyone chasing a `loss_dense_depth` shift later knows where it came from.

nvidia_tao_pytorch/cv/sparse4d/dataloader/callbacks.py
vpraveen-nv: **[minor]** Moving resampling from `on_train_batch_start` to `on_train_epoch_end` together with `reload_dataloaders_every_n_epochs=1` (train.py ~L121) and manually calling `iterator._shutdown_workers()` is a behavioural rewrite of the existing `pkl_sample_size` feature, not a co-training change. Two things to double-check: (1) `GroupInBatchSampler` is rebuilt each epoch with the same `train.seed`, so the scene-permutation sequence repeats every epoch — consider mixing `trainer.current_epoch` into the seed; (2) `_iterator` / `_shutdown_workers` are private DataLoader attributes; a comment pinning the torch version this was verified on would help the next upgrade.

nvidia_tao_pytorch/cv/sparse4d/dataloader/dataset.py
vpraveen-nv: **[minor]** Backward-compat change: `main` logged a warning for a missing `pkl_cam_counts_path` and continued; this now raises `FileNotFoundError`, raises `ValueError` if the lazy index has no embedded `pkl_cam_counts`, and (~L150) raises if `pkl_sample_size > 0 and not lazy_load`. Existing specs relying on warn-and-continue will fail at startup. Fine if intentional — please note it in the description / release notes. Also, the error text here points users at `python -m nvidia_tao_pytorch.cv.sparse4d.tools.build_lazy_index`; if the producers consolidate into tao-ds (see summary) this should name `annotations sparse4d_prepare` instead.

nvidia_tao_pytorch/cv/sparse4d/dataloader/sampler.py
vpraveen-nv: **[minor]** With `sync_route=True` and `scene_switch_iters <= 0` this only warns, yet the whole route-lockstep guarantee (each world-size block drawn from one route, ranks slicing `rank::world_size`) depends on every rank advancing scenes at the same step. Without a fixed cadence, ranks drift after the first differing group length and `_sync_cotrain_loss_keys` starts zero-filling losses every step (logged up to 20 times, then silent). Suggest raising `ValueError` here when `world_size > 1`, or auto-deriving a cadence.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
vpraveen-nv: **[question]** `_is_real_batch` requires **all** samples to be marked `has_3d_gt=False`. A mixed batch (possible with `same_scene_in_batch=False`, or when a real scene contains a few frames that do carry 3D boxes) silently falls through to the 3D loss path: the real samples contribute only background targets and the pseudo/SV losses are skipped with no log. Intended? If mixed batches are unsupported, assert homogeneity and raise so misconfiguration is visible.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_pl_model.py
vpraveen-nv: **[suggestion]** `dist.all_gather_object(gathered, sorted(losses))` pickles/broadcasts a Python list on every training step whenever `pseudo_enable` or `sv_aux_head.enable` is set. The loss-key set is a deterministic function of route × config, so you could compute the canonical key set once in `__init__` and always emit every key (zero-filled), removing the per-step collective and the desync log path. That also gives loggers a stable loss-dict shape.

nvidia_tao_pytorch/cv/sparse4d/model/instance_bank.py
vpraveen-nv: **[question]** Two questions on reset semantics: (1) the `else` branch (~L256) now calls `reset_temporal_state()` instead of `reset()`, so `prev_id` is no longer zeroed when the cache is missing / batch size changes — for test-time tracking that means instance IDs keep counting across scenes rather than restarting at 0. Does the NVSchema/HOTA path tolerate that? (2) `reset_on_time_gap` resets the **whole batch** when any single sample's `mask` is False, even though `mask` is already applied per-sample at ~L301–306. Should this be `torch.all(~mask)`, or a per-sample zeroing?

nvidia_tao_pytorch/cv/sparse4d/utils/onnx_export.py
vpraveen-nv: **[question]** This inserts `torch.where(valid, pts, 0)` into the exported graph for **all** Sparse4D exports (the exporter's static method now delegates here), so the ONNX from an unchanged 3D checkpoint differs from what `main` produces — contradicting "export unchanged". Has the resulting ONNX been run through tao-deploy `gen_trt_engine` + inference with the current MSDA TRT plugin to confirm numerical parity? Worth stating in the docstring that this relies on the plugin rejecting exactly-0 samples (`loc <= 0` returns today).

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d.py
vpraveen-nv: **[nit]** `self.cotrain_param_touch` is assigned here but never read inside `Sparse4D` (the touch lives in `Sparse4DPlModel._parameter_touch_zero`). Remove it, or drop the comment block referencing timm `checkpoint_seq` if it no longer applies.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
vpraveen-nv: **[nit]** The `clamp(-50, 50)` on classification logits also applies on the default 3D path. Numerically a no-op for healthy logits, but it silently zeroes gradients for any query whose logit exceeds the bound — another small deviation from `main` for existing runs. Fold it into the same opt-in stability flag as the non-finite handling, or document why always-on is safe.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
vpraveen-nv, line 1169: This changes the cls loss for every 3D run, not just co-training: main used `target < num_classes`, which dropped all background queries (target == C). Including them is probably right, but it's a loss-scale change hiding under "unchanged". Please call it out in the description and post before/after mAP on the release 3D checkpoint from this branch. A flag for the old behaviour would be nice but I won't block on it.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_head.py
vpraveen-nv, line 389: Main only reset the per-sample GT-index mapping here, and only under `use_temporal_align`. This now runs unconditionally and wipes the entire batch's cached state plus `dn_metas` when *any* sample changes group/scene. With `GroupInBatchSampler` slots roll over at different times, so one slot resetting truncates the temporal window for the other B-1 samples. Please keep main's per-sample reset and only do the batch-wide `reset_temporal_state()` when all samples cross a boundary (the `scene_switch_iters` case), or gate it.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
vpraveen-nv: This zeroes a NaN/Inf loss and continues, on the default 3D path. Before, the NaN was visible immediately; now a diverging run can go for hours, and `warnings.warn` only fires once per call site. Same for the `clamp(-50, 50)` and `nan_to_num(...)*0.0` fallbacks around L446/L580. Please gate these behind `train.scrub_nan_gradients` (or a new flag) so the default fails loudly like main — or at least count and raise after N consecutive non-finite steps.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_pl_model.py
vpraveen-nv, line 54: `SetCriterion` is built for evaluate/inference/export/quantize too, and with `loose_to_tight.enable=true` it now loads `mlp_ckpt` in `__init__`. So an LTT-trained model can't be evaluated or exported from its own spec unless the MLP checkpoint is present. Build the LTT/SV-aux pieces lazily on first `training_step`, or skip when not fitting.

nvidia_tao_pytorch/cv/sparse4d/dataloader/transforms.py
vpraveen-nv, line 192: Good fix (main's clip turned the -1 sentinel into a 0.1 m target), but it changes `loss_dense_depth` for existing runs. Mention it in the description / changelog.

nvidia_tao_pytorch/cv/sparse4d/dataloader/callbacks.py
vpraveen-nv, line 27: Resampling moved to epoch end + `reload_dataloaders_every_n_epochs=1` + `_shutdown_workers()` is a rewrite of the existing `pkl_sample_size` path. Two checks: `GroupInBatchSampler` is rebuilt each epoch with the same `train.seed`, so the scene permutation repeats every epoch — mix in `current_epoch`. And `_iterator`/`_shutdown_workers` are private; leave a comment with the torch version this was verified on.

nvidia_tao_pytorch/cv/sparse4d/dataloader/dataset.py
vpraveen-nv, line 642: Main warned and continued on a missing `pkl_cam_counts_path`; this raises (also for a lazy index without `pkl_cam_counts`, and `pkl_sample_size > 0` without `lazy_load`). Fine, but note it in the description. Also the error text should mention `annotations sparse4d_prepare` (tao-ds) rather than only the `tools/` module.

nvidia_tao_pytorch/cv/sparse4d/dataloader/sampler.py
vpraveen-nv, line 264: With `sync_route=True` and `scene_switch_iters <= 0` this only warns, but route lockstep across ranks depends on a fixed cadence. Without it ranks drift after the first differing group length and `_sync_cotrain_loss_keys` zero-fills every step. Raise here when `world_size > 1`.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
vpraveen-nv: `_is_real_batch` needs *all* samples to have `has_3d_gt=False`. A mixed batch silently takes the 3D path — real samples become background-only and pseudo/SV losses are skipped with no log. Intended? If mixed batches aren't supported, assert and raise.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_pl_model.py
vpraveen-nv: `all_gather_object` on every step whenever pseudo/SV is enabled. The loss-key set is fixed per route × config, so you can compute it once in `__init__` and always emit every key zero-filled — no per-step collective, no desync path, stable dict shape for loggers.

nvidia_tao_pytorch/cv/sparse4d/model/instance_bank.py
vpraveen-nv: Two questions: the `else` branch (~L256) now calls `reset_temporal_state()` instead of `reset()`, so `prev_id` isn't zeroed on cache miss — instance IDs keep counting across scenes at test time. Does the HOTA/NVSchema path care? And `reset_on_time_gap` resets the whole batch if any one sample's `mask` is False, though `mask` is already applied per-sample below. Should that be `all(~mask)`?

nvidia_tao_pytorch/cv/sparse4d/utils/onnx_export.py
vpraveen-nv, line 18: This adds `torch.where(valid, pts, 0)` to every Sparse4D export, so the ONNX from an unchanged 3D checkpoint differs from main. Has it been through tao-deploy `gen_trt_engine` + inference with the current MSDA plugin? Also worth a docstring note that this relies on the plugin skipping exactly-0 samples.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d.py
vpraveen-nv: `self.cotrain_param_touch` is never read in `Sparse4D` (the touch lives in `Sparse4DPlModel`). Drop it or the comment.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
vpraveen-nv: The ±50 clamp also runs on the default path and zeroes gradients past the bound. Fold into the same opt-in flag as the non-finite handling.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
yizwang-nv: Partially addressed in 2f8f99519d4d89126e8ecf07f3da7cb5376deeea. The compatibility notes now explicitly state that background queries (target == C) contribute to focal loss, changing the classification-loss scale and gradients for existing 3D training. The updated PR description is prepared with the same caveat.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_head.py
yizwang-nv: Addressed in 2f8f99519d4d89126e8ecf07f3da7cb5376deeea. Partial group/scene boundaries now reset only affected GT-index mappings, and only under use_temporal_align. Full temporal state and DN caches are cleared only when all batch slots cross a boundary. Tests cover group and scene changes, partial and all-slot boundaries, and temporal alignment enabled/disabled.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
yizwang-nv: Addressed in 2f8f99519d4d89126e8ecf07f3da7cb5376deeea. Default training now raises `FloatingPointError` on non-finite losses. Classification-logit clipping, sanitization of graph-zero inputs, and non-finite loss replacement require `train.scrub_nan_gradients=true`. Main, DN, and pseudo-label classification paths follow this policy. Valid empty positive sets return graph-connected scalar zeros instead of creating mean(empty) NaNs. Added default-failure, opt-in recovery, gradient, and empty-target regression tests.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_pl_model.py
yizwang-nv: Addressed in 2f8f99519d4d89126e8ecf07f3da7cb5376deeea. Evaluation, inference, export, and quantization no longer build `SetCriterion`, so an LTT-enabled saved spec does not require its training-only MLP. The training CLI explicitly passes `build_training_losses=True` at construction, before checkpoint restoration and optimizer/DDP setup; trainable auxiliary parameters are not created lazily after setup. Tests cover non-training construction with an absent MLP and restoration of registered criterion parameters.

nvidia_tao_pytorch/cv/sparse4d/dataloader/transforms.py
yizwang-nv: Addressed in 2f8f99519d4d89126e8ecf07f3da7cb5376deeea. The compatibility notes now explicitly describe the -1 sentinel correction and its effect on existing loss_dense_depth values and gradients: invalid pixels no longer become artificial 0.1 m targets. The same note is included in the prepared PR-description update.

nvidia_tao_pytorch/cv/sparse4d/dataloader/callbacks.py
yizwang-nv: Addressed in 2f8f9951.
- Rebuilt samplers now use train.seed + trainer.current_epoch, so scene ordering varies across epochs while remaining reproducible for the same seed and epoch.
- Added a comment documenting that _iterator / _shutdown_workers() were verified against TAO’s PyTorch 2.11.0a0 nv26.03 runtime.
Added passing regression tests for same-epoch reproducibility, changed ordering across epochs, and actual persistent-worker shutdown. The worker test also verifies that a replacement iterator observes the newly resampled dataset.

nvidia_tao_pytorch/cv/sparse4d/dataloader/sampler.py
yizwang-nv: Updated in 2f8f9951. The compatibility notes explicitly document this behavior change: PKL subset sampling requires lazy_load=True and complete camera counts; a configured-but-missing pkl_cam_counts_path now raises instead of warning and continuing.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
yizwang-nv: Addressed in `2f8f9951`. _is_real_batch now rejects mixed 2D/3D supervision within a batch with a clear ValueError directing callers to route-homogeneous same-scene sampling. Tests cover list/tensor flags and ambiguous missing flags beside explicitly 2D samples.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d_pl_model.py
yizwang-nv: Addressed in `2f8f9951`. The criterion builds a configuration-derived loss-key schema once during construction. The Lightning module emits that stable schema with graph-connected zeros for missing route losses, without `all_gather_object` or another per-step key collective. Unknown keys raise rather than being silently dropped.

nvidia_tao_pytorch/cv/sparse4d/model/instance_bank.py
yizwang-nv: 1. Preserving `prev_id` is intentional: temporal resets clear cached associations without recycling IDs within an inference run. NVSchema preserves IDs as strings; HOTA preprocessing remaps them contiguously per sequence, so a constant ID offset does not change the score. Added tests using the actual NVSchema converter and HOTA preprocessing/scoring to verify this. InstanceBank.reset() still resets the counter for a new run. One caveat: HOTA’s lookup allocation scales with the maximum numeric ID, so independent runs should use that explicit reset.
2. Agreed—the whole-batch reset now occurs only when all samples are invalid, implemented as not torch.any(mask), equivalent to torch.all(~mask). Partial gaps retain the existing per-sample masking behavior. Regression tests cover both partial and all-invalid batches.

nvidia_tao_pytorch/cv/sparse4d/utils/onnx_export.py
yizwang-nv: Partially addressed in `2f8f9951`. The export helper now explicitly states that the deployment MSDA plugin must reject exactly-zero coordinates using strict (0, 1) bounds, and that this applies to exports of unchanged 3D checkpoints. ONNX export tests pass. Working on NVIDIA-TAO/tao-deploy Sparse4D backend.

nvidia_tao_pytorch/cv/sparse4d/model/sparse4d.py
yizwang-nv: Addressed in `2f8f9951`. Removed the unused Sparse4D.cotrain_param_touch attribute. Parameter touching remains owned by Sparse4DPlModel, where it is actually used.

nvidia_tao_pytorch/cv/sparse4d/model/criterion.py
yizwang-nv: Addressed in `2f8f9951`. The main, DN, and pseudo-label classification paths now clamp logits only when `train.scrub_nan_gradients=true`. A regression checks that finite extreme logits retain non-zero focal-loss gradients by default and are clipped only in the opt-in mode.

nvidia_tao_pytorch/cv/sparse4d/dataloader/dataset.py
yizwang-nv: This is an intentional change from warning-and-continuing to fail-fast validation. I’ll document these compatibility changes in the PR description:
- A configured `pkl_cam_counts_path` that does not exist now raises.
- A lazy index missing `pkl_cam_counts` now raises.
- `pkl_sample_size > 0` requires `lazy_load=True`.
I’ll also update the error messages to recommend regenerating the preparation artifacts with annotations sparse4d_prepare in tao-data-services, rather than referring only to the tools/ module.

